PR #7, opened 2015-06-10
Make SmallVec::drop sound and other stuff

SimonSapin:
* Use an enum so that we’re not dropping uninitialized `T` inside `[T; $size]`
* Move fields into that enum to make the struct smaller
* Refactor to be generic over array types rather than generate multiple `SmallVecN` types with macros.

r? @Manishearth

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,3 +11,6 @@ readme = "README.md"
 name = "smallvec"
 path = "lib.rs"
 doctest = false
+
+[dependencies]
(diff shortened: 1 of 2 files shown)


Manishearth:
r+

bluss:
With acknowledgment of: using ideas like Nodrop & Array trait from @bluss' arrayvec crate.

SimonSapin:
Yes, sorry I forgot to mention.

PR #15, opened 2015-08-20
Fix panic safety issue in drop

### LINKED ISSUE #14: Panic safety in drop
@stebalien discovered panic safety issue bluss/arrayvec#3 and it applies to smallvec as well.

My understanding is: SmallVec::drop first attempts to drop every element, *then* it inhibits the drop of the inner array. The panic safety issue is that a panic during drop of an element means the inhibition is never reached, so the inner data can be dropped again.

Testcase adapted from @stebalien

```rust
    #[test]
    #[should_panic]
    fn test_drop_panic_smallvec() {
        // This test should only panic once, and not double panic,
        // which would mean a double drop
        struct DropPanic;

        impl Drop for DropPanic {
            fn drop(&mut self) {
                panic!("drop");
            }
        }

        let mut v = SmallVec::<[_; 1]>::new();
        v.push(DropPanic);
    }
```

bluss:
The summary is: SmallVec::drop first attempts to drop every element,
then it inhibits the drop of the inner array. The panic safety issue is
that a panic during drop of an element means the inhibition is never
reached, so the inner data can be dropped again.

If Drop is split betweeen SmallVec and SmallVecData, this issue is
avoided because the SmallVecData drop will be called even in the panic
case.

This solution incurs the overhead of an additional drop flag on
SmallVecData.

Fixes #14

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -83,6 +83,23 @@ enum SmallVecData<A: Array> {
     Heap { ptr: *mut A::Item, capacity: usize },
 }
 
+impl<A: Array> Drop for SmallVecData<A> {
+    fn drop(&mut self) {
+        unsafe {
+            match *self {
+                ref mut inline @ Inline { .. } => {
+                    // Inhibit the array destructor.
+                    ptr::write(inline, Heap {
+                        ptr: ptr::null_mut(),
+                        capacity: 0,
+                    });
+                }
+                Heap { ptr, capacity } => deallocate(ptr, capacity),
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


bluss:
This may not be the solution you want. But it should help anyway, to illustrate the problem(?). Welcome to use the test if not the fix.

bluss:
Updated fix to not use NoDrop -- the inner SmallVecData serves the same purpose.

SimonSapin:
Yes, I prefer this second version. Thanks!

PR #20, opened 2016-03-30
Implement Send and Sync on SmallVec

tomaka:
Because of the core team's questionable decision of not implementing Send and Sync on raw pointers, we need to implement them manually.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/20)


pcwalton:
@bors-servo: r+

For what it's worth, this was not a core team decision

PR #22, opened 2016-05-24
Implement PartialEq, Eq, PartialOrd, Ord on SmallVec.

cfallin:
Currently one can compare slices, but having the traits on the `SmallVec` itself is slightly more convenient. Perhaps there's a reason these traits aren't yet implemented, but if not, here's a PR!

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/22)


SimonSapin:
Looks good, thanks!

@bors-servo r+

PR #26, opened 2016-07-31
Bring SmallVec closer to Vec

Marwes:
This PR implements almost all traits which [Vec](https://doc.rust-lang.org/std/vec/struct.Vec.html) implements for `SmallVec`, bringing it closer to being a drop in replacement for `Vec`. A few implementations were omitted such as `Extend<&A::Item>` which could not be implemented due to a conflict with `Extend<A::Item>`, a few `From` implementations which may be a bit more esoteric (easily added if so desired) and `Write`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/26)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -935,4 +935,69 @@ pub mod tests {
             assert_eq!(a.hash(&mut hasher), b.hash(&mut hasher));
         }
     }
+
+    #[test]
+    fn test_as_ref() {
+        let mut a: SmallVec<[u32; 2]> = SmallVec::new();
+        a.push(1);
+        assert_eq!(a.as_ref(), [1]);
+        a.push(2);
+        assert_eq!(a.as_ref(), [1, 2]);
+        a.push(3);
+        assert_eq!(a.as_ref(), [1, 2, 3]);
+    }
+
+    #[test]
+    fn test_as_mut() {
+        let mut a: SmallVec<[u32; 2]> = SmallVec::new();
+        a.push(1);
+        assert_eq!(a.as_mut(), [1]);
+        a.push(2);
+        assert_eq!(a.as_mut(), [1, 2]);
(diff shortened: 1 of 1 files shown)


Marwes:
The renaming of `into_iter` to `drain` is a breaking change but was necessary due to implementing `IntoIter` for `SmallVec` (so it can be used directly in for loops). `drain` is also not an exact match for `Vec::drain` as [RangeArgument](https://doc.rust-lang.org/collections/range/trait.RangeArgument.html) is not yet stabilized.

Ms2ger:
Might be worth adding tests for borrow/as_ref and anything else I missed?

Marwes:
Added tests for traits that missed them!

Marwes:
Bump

lib.rs
```diff
@@ -443,8 +501,102 @@ impl<A: Array> Ord for SmallVec<A> where A::Item: Ord {
...
+    fn drop(&mut self) {
+        for _ in self { }
+    }
+}
+
+impl<A: Array> Iterator for IntoIter<A> {
+    type Item = A::Item;
+    
+    #[inline]
+    #[inline]
```
rphmeier: two inline annotations?

Marwes: fixed

rphmeier:
Looks good to me; I'd appreciate if this could land soon :)

SimonSapin:
Looks good, thanks!

@bors-servo r+

SimonSapin:
http://build.servo.org/homu/queue/rust-smallvec shows this PR as approved, but it doesn’t look like a Travis build is being scheduled. CC @larsbergstrom 

https://travis-ci.org/servo/rust-smallvec/builds/151422387 is green.

PR #27, opened 2016-08-19
Support std::iter::extend<T> for VecLike<T>

nipunn1313:
VecLike naturally supports push, so it also makes sense to support extend. Both implementors SmallVec and Vec already have tested implementations, so this extension feels natural.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/27)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -30,7 +30,8 @@ pub trait VecLike<T>:
         ops::Index<ops::RangeFull, Output=[T]> +
         ops::IndexMut<ops::RangeFull> +
         ops::Deref +
-        ops::DerefMut {
+        ops::DerefMut +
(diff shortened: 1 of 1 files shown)


nox:
Superseded by https://github.com/servo/rust-smallvec/pull/26.

rphmeier:
(not fully -- this PR also adds an `Extend<T>` bound on `VecLike`)

nox:
Oh right. Could you rebase it?

SimonSapin:
Adding a bound is a breaking change for implementors. Please change the version number accordingly.

nipunn1313:
Rebased! I'll leave it to you to update the version number.

PR #29, opened 2016-08-23
Implement extend_from_slice and insert_from_slice with memmove optimization

nipunn1313:
Implement the performance optimized versions of insert_many and extend (from #28). These methods use memmove rather than a looped insert.

If we had function specialization, we could implement these without exposing new methods. Up to the maintainers whether we want to support these new methods.

---

This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/29)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -477,6 +477,29 @@ impl<A: Array> SmallVec<A> {
     }
 }
 
+impl<A: Array> SmallVec<A> where A::Item: Copy {
+    pub fn insert_slice(&mut self, index: usize, slice: &[A::Item]) {
+        self.reserve(slice.len());
+
+        let len = self.len;
+        assert!(index <= len);
+
+        unsafe {
+            let slice_ptr = slice.as_ptr();
+            let ptr = self.as_mut_ptr().offset(index as isize);
+            ptr::copy(ptr, ptr.offset(slice.len() as isize), len - index);
+            ptr::copy(slice_ptr, ptr, slice.len());
+            self.set_len(len + slice.len());
+        }
+    }
+
+    #[inline]
+    pub fn extend_slice(&mut self, slice: &[A::Item]) {
+        let len = self.len();
+        self.insert_slice(len, slice);
+    }
+}
+
 impl<A: Array> ops::Deref for SmallVec<A> {
     type Target = [A::Item];
     #[inline]
@@ -1066,6 +1089,28 @@ pub mod tests {
         let mut v: SmallVec<[u8; 8]> = SmallVec::new();
         v.extend(0..8);
         v.grow(5);
+    }
+
+    #[test]
+    fn test_insert_slice() {
+        let mut v: SmallVec<[u8; 8]> = SmallVec::new();
+        for x in 0..4 {
+            v.push(x);
+        }
+        assert_eq!(v.len(), 4);
+        v.insert_slice(1, &[5, 6]);
+        assert_eq!(&v.iter().map(|v| *v).collect::<Vec<_>>(), &[0, 5, 6, 1, 2, 3]);
+    }
+
+    #[test]
+    fn test_extend_slice() {
+        let mut v: SmallVec<[u8; 8]> = SmallVec::new();
+        for x in 0..4 {
+            v.push(x);
+        }
+        assert_eq!(v.len(), 4);
+        v.extend_slice(&[5, 6]);
+        assert_eq!(&v.iter().map(|v| *v).collect::<Vec<_>>(), &[0, 1, 2, 3, 5, 6]);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


Marwes:
Might want to call `extend_slice`, [copy_from_slice](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.copy_from_slice) to mimic the `method on `Vec`

lib.rs
```diff
@@ -315,6 +315,29 @@ impl<A: Array> SmallVec<A> {
     }
 }
 
+impl<A: Array> SmallVec<A> where A::Item: Copy {
+    pub fn insert_slice(&mut self, index: usize, slice: &[A::Item]) {
+        self.reserve(slice.len());
```
emilio: Hmm... I'm kind of sleepy right now, but I think here you should reserve `index + slice.len()`?

nipunn1313: I believe reserve takes in an "additional" amount required and ensures the capacity is large enough, so I think this is ok.

nipunn1313:
I think "extend_from_slice" and "insert_from_slice" are also reasonable names. I think "copy_from_slice" is misleading because it's not overwriting existing contents.

Marwes:
[extend_from_slice](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend_from_slice) is also a method on `Vec` except it takes `&[T] where T: Clone`.

lib.rs
emilio: Oh, you're right, nvm, I thought `reserve()` did the usual stuff, that is, ensuring capacity.

nipunn1313:
That makes sense. I agree with that.

nipunn1313:
I went ahead and rebased this on master. Here are the benchmarks on my local machine

```
Nipunns-MacBook-Pro:rust-smallvec nipunn$ cargo bench bench
    Finished release [optimized] target(s) in 0.0 secs
     Running target/release/bench-760919c65e248f9d

running 7 tests
test bench_extend            ... bench:         149 ns/iter (+/- 26)
test bench_extend_from_slice ... bench:          65 ns/iter (+/- 8)
test bench_insert            ... bench:       1,411 ns/iter (+/- 440)
test bench_insert_from_slice ... bench:         119 ns/iter (+/- 22)
test bench_insert_many       ... bench:         483 ns/iter (+/- 59)
test bench_push              ... bench:         425 ns/iter (+/- 217)
test bench_pushpop           ... bench:         351 ns/iter (+/- 78)

test result: ok. 0 passed; 0 failed; 0 ignored; 7 measured

     Running target/release/deps/smallvec-82287e02aa187b8d

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured

Nipunns-MacBook-Pro:rust-smallvec nipunn$ git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
Nipunns-MacBook-Pro:rust-smallvec nipunn$ cargo bench bench
   Compiling smallvec v0.3.1 (file:///Users/nipunn/src/rust-smallvec)
    Finished release [optimized] target(s) in 2.54 secs
     Running target/release/bench-760919c65e248f9d

running 5 tests
test bench_extend      ... bench:         151 ns/iter (+/- 16)
test bench_insert      ... bench:       1,407 ns/iter (+/- 369)
test bench_insert_many ... bench:         480 ns/iter (+/- 75)
test bench_push        ... bench:         430 ns/iter (+/- 152)
test bench_pushpop     ... bench:         353 ns/iter (+/- 104)

test result: ok. 0 passed; 0 failed; 0 ignored; 5 measured

     Running target/release/deps/smallvec-82287e02aa187b8d

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured
```

Looks like there shouldn't be regressions, though we should be extra careful to make sure there aren't integer overflow / memory issues.

jdm:
@bors-servo: r+
Thanks!

nox:
Why does that use `Copy` instead of `Clone` like the stdlib?

nipunn1313:
unsafe ptr::copy will only work if the type is Copy.
If the type is clone, then we have to iterate over the elements and run the clone code.
You're right that it's a little bit different than the stdlib which is unfortunate.

If function specialization lands (https://github.com/rust-lang/rust/issues/31844), then this naming / perf tradeoff may become moot.

nox:
The stdlib uses `Clone` but manages to optimise it nonetheless without specialisation AFAIK. Just because you iterate and clone doesn't mean it has to be slow.

nox:
Ah no, nevermind: https://github.com/rust-lang/rust/blob/master/src/libcollections/vec.rs#L1691-L1703

nipunn1313:
Interesting! That looks like it uses specialization. At the time we made
this PR, specialization wasn't available.
--Nipunn

PR #30, opened 2016-08-24
Implement Default for SmallVec

Marwes:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/30)


Marwes:
Missed adding Default in my previous PR.

PR #31, opened 2016-08-27
Add benchmarks for push, insert, extend, and pushpop

nipunn1313:
Want to add these benchmarks so that we can more effectively detect regressions/improvements from PR's like #28 and #29.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/31)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -5,6 +5,8 @@
 //! Small vectors in various sizes. These store a certain number of elements inline and fall back
 //! to the heap for larger allocations.
 
+#![feature(test)]
+
 use std::borrow::{Borrow, BorrowMut};
 use std::cmp;
 use std::fmt;
@@ -647,8 +649,10 @@ impl_array!(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 20, 24, 32,
 
 #[cfg(test)]
 pub mod tests {
+    extern crate test;
     use SmallVec;
     use std::borrow::ToOwned;
+    use self::test::Bencher;
 
     // We heap allocate all these strings so that double frees will show up under valgrind.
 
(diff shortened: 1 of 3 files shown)


lib.rs
```diff
@@ -5,6 +5,8 @@
 //! Small vectors in various sizes. These store a certain number of elements inline and fall back
 //! to the heap for larger allocations.
 
+#![feature(test)]
```
emilio: This won't compile in stable, will it?

lib.rs
```diff
@@ -1000,4 +1004,43 @@ pub mod tests {
         assert_eq!(vec.clone().into_iter().len(), 3);
         assert_eq!(vec.drain().len(), 3);
     }
+
+    #[bench]
+    fn bench_push(b: &mut Bencher) {
+        b.iter(|| {
+            let mut vec: SmallVec<[u64; 16]> = SmallVec::new();
+            for x in 0..100 {
```
emilio: Also, I think the compiler will be pretty much able to optimize these loops out. IIRC there was some builtin support for black box testing for this kind of things. I think we should be able to enforce that using it, or at least confirm it's not being optimized out.

nipunn1313: I believe if the closure returns a value then the optimizer will not remove it.

I was empirically able to determine that the optimizer was not removing the code either way (for now), but I will include the vec return value to ensure it.

lib.rs
nipunn1313: Correct. Let me work on factoring it into a feature which is disabled by default.

nipunn1313:
Looks like this on my machine (Mac book pro laptop)
```
Nipunns-MacBook-Pro:rust-smallvec nipunn$ cargo bench --verbose --features=benchmarks bench
       Fresh smallvec v0.2.0 (file:///Users/nipunn/src/rust-smallvec)
    Finished release [optimized] target(s) in 0.0 secs
     Running `/Users/nipunn/src/rust-smallvec/target/release/deps/smallvec-a59689f47ae774cb bench --bench`

running 4 tests
test bench::bench_extend  ... bench:         169 ns/iter (+/- 19)
test bench::bench_insert  ... bench:       1,201 ns/iter (+/- 267)
test bench::bench_push    ... bench:         295 ns/iter (+/- 39)
test bench::bench_pushpop ... bench:         357 ns/iter (+/- 117)

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured
```

Travis CI looked like this. I don't think we can really rely on it though.
```
running 4 tests
test bench::bench_extend  ... bench:         182 ns/iter (+/- 3)
test bench::bench_insert  ... bench:       1,916 ns/iter (+/- 21)
test bench::bench_push    ... bench:         307 ns/iter (+/- 6)
test bench::bench_pushpop ... bench:         383 ns/iter (+/- 6)```

nipunn1313:
Any update on this?

jdm:
@bors-servo: r+
Thanks for doing this work!

PR #35, opened 2016-10-03
Remove unnecessary `VecLike` trait.

frewsxcv:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/35)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -482,11 +482,11 @@ impl<A: Array> Drop for IntoIter<A> {
 
 impl<A: Array> Iterator for IntoIter<A> {
     type Item = A::Item;
-    
+
     #[inline]
     fn next(&mut self) -> Option<A::Item> {
         if self.current == self.end {
-            None    
+            None
         }
         else {
             unsafe {
@@ -508,7 +508,7 @@ impl<A: Array> DoubleEndedIterator for IntoIter<A> {
     #[inline]
     fn next_back(&mut self) -> Option<A::Item> {
         if self.current == self.end {
-            None    
+            None
         }
         else {
             unsafe {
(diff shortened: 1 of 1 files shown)


frewsxcv:
It's not clear to me that this `trait` serves any purpose. Am I overlooking something?

lib.rs
```diff
@@ -150,12 +119,6 @@ impl<A: Array> SmallVec<A> {
     pub fn inline_size(&self) -> usize {
         A::size()
     }
-    pub fn len(&self) -> usize {
-        self.len
-    }
-    pub fn is_empty(&self) -> bool {
-        self.len == 0
-    }
```
frewsxcv: These methods are available via `Deref`, right?

mbrubeck:
This is used to write generic code that works with both `Vec<T>` and `SmallVec<T>`. It's used in Servo's `style` crate, for example. Closing because we can't remove this without breaking existing code.

frewsxcv:
via IRC:

```
19:07 <ghservo> [rust-smallvec] frewsxcv opened pull request #35: Remove unnecessary `VecLike` trait. (master...rm-veclike) https://git.io/vPOao
19:10 <frewsxcv> Corey Someone want to sanity check (and/or review) that PR? ^
19:11 <•mbrubeck> frewsxcv: VecLike is used in the style crate
19:11 <frewsxcv> Corey Maybe it should live in the style crate?
19:11 Good to know though
19:12 <•mbrubeck> frewsxcv: It lets you write code that's generic over Vec<T> and SmallVec<T>, so it seems generally useful
19:15 <frewsxcv> Corey Fair enough.
```

Ms2ger:
I guess it would be fine to move to style.

mbrubeck:
We may want to revisit this, since the style crate no longer needs this (servo/servo#19424), and code that does use it could be converted to use standard traits like `Extend` instead.

PR #38, opened 2016-10-05
Specify that VecLike derefs to a slice

Marwes:
I believe this is just an oversight. Since `<[T]>::len` is now available through the `Deref` bound I also removed `VecLike::len`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/38)


jdm:
@bors-servo: r+
Thanks!

PR #42, opened 2017-03-14
Add HeapSizeOf trait and from_slice method

nipunn1313:
Also added some tests for the new methods. The rationale
- from_slice is an ergonomic way to convert a Copy-able slice to a SmallVec (instead of using the From<'a slice> which uses an iterator)
- HeapSizeOf is handy for measuring heap allocations. Especially useful for monitoring something like this. Seems to be part of the servo project anyway.
- Added size 36 to smallvec sizes (was useful to us)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/42)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -6,16 +6,20 @@
 //! to the heap for larger allocations.  This can be a useful optimization for improving cache
 //! locality and reducing allocator traffic for workloads that fit within the inline buffer.
 
+extern crate heapsize;
+
 use std::borrow::{Borrow, BorrowMut};
 use std::cmp;
 use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::iter::{IntoIterator, FromIterator};
 use std::mem;
+use std::os::raw::c_void;
 use std::ops;
 use std::ptr;
 use std::slice;
 
+use heapsize::{HeapSizeOf, heap_size_of};
 use SmallVecData::{Inline, Heap};
 
 /// Common operations implemented by both `Vec` and `SmallVec`.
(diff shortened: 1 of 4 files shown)


lib.rs
```diff
@@ -478,6 +482,12 @@ impl<A: Array> SmallVec<A> {
 }
 
 impl<A: Array> SmallVec<A> where A::Item: Copy {
+    pub fn from_slice(slice: &[A::Item]) -> Self {
```
rphmeier: Could it just be `impl<A: Array, T: AsRef<[A::Item]>> From<T> for SmallVec<A>`?

nipunn1313: Also need to make sure A::Item: Copy which is what made me make it more specific (from_slice() rather than impl From).

I'd imagine that a generic From implementation should work regardless of whether it's Copy or not. Without specialization, it's hard to make an efficient version for the Copy type, so I went with a different "from_slice" function.

tomaka:
For the record `heapsize` is a public external dependency and will may a dependency hell later.

rphmeier:
@tomaka That can be addressed by putting heapsize stuff under a feature flag

jdm:
Let's have an optional feature flag for heapsize, as mentioned in previous comments. Otherwise these changes look unobjectionable to me.

lib.rs
```diff
@@ -6,6 +6,9 @@
 //! to the heap for larger allocations.  This can be a useful optimization for improving cache
 //! locality and reducing allocator traffic for workloads that fit within the inline buffer.
 
+#![feature(specialization)]
```
jdm: This looks like a mistake?

nipunn1313: Good catch. I was playing around with specialization and accidentally rolled it into this. Fixed!

nipunn1313:
Thanks yo!

PR #43, opened 2017-03-22
Don't zero the array on initialization

mbrubeck:
Using `mem::uninitialized` is potentially faster than `mem::zeroed`, and doesn't remove any safety (since both uninitialized memory and zeroed memory are unsafe to read from in general).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/43)


mbrubeck:
@bors-servo retry

* #44

Issue #44, opened 2017-03-22
Remove benchmark feature in stable builds


jdm:
From travis:
```
     Running `rustc --crate-name bench benches/bench.rs --emit=dep-info,link -C opt-level=3 --test -C metadata=6041378d75069d53 -C extra-filename=-6041378d75069d53 --out-dir /home/travis/build/servo/rust-smallvec/target/release/deps -L dependency=/home/travis/build/servo/rust-smallvec/target/release/deps --extern smallvec=/home/travis/build/servo/rust-smallvec/target/release/deps/libsmallvec-ad2736f846965482.rlib`

error[E0554]: #[feature] may not be used on the stable release channel

 --> benches/bench.rs:1:1

  |

1 | #![feature(test)]

  | ^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: Could not compile `smallvec`.

Caused by:

  process didn't exit successfully: `rustc --crate-name bench benches/bench.rs --emit=dep-info,link -C opt-level=3 --test -C metadata=6041378d75069d53 -C extra-filename=-6041378d75069d53 --out-dir /home/travis/build/servo/rust-smallvec/target/release/deps -L dependency=/home/travis/build/servo/rust-smallvec/target/release/deps --extern smallvec=/home/travis/build/servo/rust-smallvec/target/release/deps/libsmallvec-ad2736f846965482.rlib` (exit code: 101)

Build failed, waiting for other jobs to finish...

error: build failed

The command "cargo build --verbose &&

cargo build --features=heapsizeof --verbose &&

cargo test --verbose &&

cargo test --features=heapsizeof --verbose &&

([ $TRAVIS_RUST_VERSION != nightly ] || cargo bench --verbose bench)

" exited with 101.
```

mbrubeck:
The travis script should already do this, but its check seems to be broken, maybe related to travis-ci/travis-build#993 which just landed.

mbrubeck:
I temporarily disabled `cargo bench`.  We should re-enable it after the Travis issue is fixed.

mbrubeck:
Fixed by travis-ci/travis-ci#7510.

PR #45, opened 2017-03-22
Add SmallVec::into_vec

mbrubeck:
Converts a SmallVec to a Vec, without reallocating if the SmallVec has already spilled onto the heap.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/45)


emilio:
@bors-servo r+

Though travis seems to fail...

mbrubeck:
See #44 for the failure.

PR #48, opened 2017-03-29
Add from_vec method for efficient creation from a Vec.

SergioBenitez:
I also bumped the version to 0.3.3 for easy publishing.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/48)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "smallvec"
-version = "0.3.2"
+version = "0.3.3"
 authors = ["Simon Sapin <simon.sapin@exyr.org>"]
 license = "MPL-2.0"
 repository = "https://github.com/servo/rust-smallvec"
--- a/lib.rs
+++ b/lib.rs
@@ -1388,4 +1388,37 @@ pub mod tests {
         let vec = SmallVec::<[u8; 2]>::from_iter(0..3);
         assert_eq!(vec.into_vec(), vec![0, 1, 2]);
     }
+
+    #[test]
+    fn test_from_vec() {
+        let vec = vec![];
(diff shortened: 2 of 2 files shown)


emilio:
Could you add a few tests for this? In particular, an empty vector pointer may not be an actual valid pointer IIRC (may be `0x1`?).

SergioBenitez:
Sure, added.

emilio:
Oh, ok, so we don't rely on the validness of the pointer (we just rely on `Vec` for most of stuff). r=me then

@bors-servo r+

Thanks for doing this!

SergioBenitez:
Well, we certainly take the raw pointer from `Vec`, but we do so via the public API. The documentation doesn't say anything about the pointer possibly being invalid.

emilio:
> Well, we certainly take the raw pointer from Vec, but we do so via the public API. The documentation doesn't say anything about the pointer possibly being invalid.

Right, there's no guarantee about that AFAIK. See https://bugzilla.mozilla.org/show_bug.cgi?id=1344209, which is the reason I was wary of this.

SergioBenitez:
Looking at the source code for `Vec`, `ptr` will indeed be some constant, invalid value when the `Vec` is empty (`alloc::heap::EMPTY`). However, the ptr will only be this value when capacity/length are zero or `T` is a zero-sized type. As a result, `SmallVec` should never dereference the invalid pointer, and everything should work exactly as expected.

In the bug you linked, it seems that the fix was to use a `null` pointer when the string is empty. Presumably this is because some of the Core Foundation/Foundation code checks the `ptr` value for nullness to determine if it can be dereferenced. None of the code in `SmallVec` does this, another indication that everything is alright.

emilio:
Right, it's not about core foundation, but Gecko's code, but that's right :)

PR #49, opened 2017-05-20
Add no_std support

Pratyush:
This library can easily support `no_std` code on `nightly`; it does require the `collections` feature, however. This PR adds support for this feature by enabling a on-by-default `std` feature. This feature can be turned off to support `no_std` mode.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/49)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,10 @@ readme = "README.md"
 documentation = "http://doc.servo.org/smallvec/"
 
 [features]
-heapsizeof = ["heapsize"]
+heapsizeof = ["heapsize", "std"]
+collections = []
+std = []
+default = ["std"]
 
 [lib]
 name = "smallvec"
--- a/lib.rs
+++ b/lib.rs
@@ -6,8 +6,23 @@
 //! to the heap for larger allocations.  This can be a useful optimization for improving cache
 //! locality and reducing allocator traffic for workloads that fit within the inline buffer.
 
+#![cfg_attr(not(feature = "std"), no_std)]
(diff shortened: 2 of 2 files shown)


nox:
Usually crates define a `no_std` method, not a default `std` one.

nox:
Or maybe I dreamt that... I was pretty sure it was the case for `serde` but turns out it's not. Disregard me.

Pratyush:
Sure, I can do that too. Some crates follow that convention, but some others don't (like `serde`). It's a bit of a mess in that regard. I followed the convention that `serde` is using.

nox:
Following `serde` is always a good default choice. :)

est31:
no_std is an anti pattern, see https://github.com/rust-lang/rfcs/pull/1841 (The RFC itself is being rejected, but it stays an antipattern).

mbrubeck:
I think `no_std` support is definitely a good thing for this crate, and I don't see any remaining questions about this patch.  I plan to r+ it unless there are any objections.

Unfortunately we delayed long enough that the `collections` crate has been deprecated and its contents have been moved to `alloc`.  Could you update this to use `alloc`, please?

mbrubeck:
@bors-servo r+

Since the move to `alloc` only causes a warning for now, I think it's easier to merge this and then file a follow-up for that issue.

Pratyush:
The other thing is that I believe that the collections crate still exists; now it's just a wrapper around `alloc`: https://github.com/rust-lang/rust/pull/42720

PR #52, opened 2017-06-01
impl std::io::Write for SmallVec

chpio:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/52)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -33,6 +33,8 @@ use std::mem;
 use std::ops;
 use std::ptr;
 use std::slice;
+#[cfg(feature = "std")]
+use std::io;
 #[cfg(feature="heapsizeof")]
 use std::os::raw::c_void;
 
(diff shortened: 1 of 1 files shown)


chpio:
this collides a little bit with the [*no_std* pr](https://github.com/servo/rust-smallvec/pull/49), because all other *std* modules are part of *core* but *io* is not.

mbrubeck:
#49 has landed now, so this should be rebased on top of it, and `#[cfg(feature = "std")]` should be added to all the new code.

lib.rs
```diff
@@ -1436,4 +1458,21 @@ pub mod tests {
         assert_eq!(&*small_vec, &[1, 2, 3, 4, 5]);
         drop(small_vec);
     }
+
+    #[cfg(feature = "std")]
```
chpio: Is that the correct way of doing this?

mbrubeck:
@bors-servo r+

Yes, thanks!  When merged with #53, this builds and passes tests both with and without the `std` feature.

PR #53, opened 2017-06-20
Document and test no_std support

mbrubeck:
This builds on the new "std" Cargo feature added in #49.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/53)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "smallvec"
-version = "0.4.0"
+version = "0.4.1"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -11,7 +11,6 @@ documentation = "http://doc.servo.org/smallvec/"
 
 [features]
 heapsizeof = ["heapsize", "std"]
-collections = []
```
mbrubeck: This was unused.

PR #54, opened 2017-06-27
Add ExtendFromSlice trait

Vurich:
At Parity we tried to switch from our internal `elastic-array` crate (which has soundness and ergonomics issues) to `smallvec` but the PR that attempted this replaced `extend_from_slice` calls with `push` loops (which are much slower). We could fix this internally, but the discussion around it made me wonder if this would be something that could be useful upstream.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/54)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -88,6 +88,36 @@ impl<T> VecLike<T> for Vec<T> {
     }
 }
 
+/// Trait to be implemented by a collection that can be extended from a slice
+///
+/// ## Example
+///
+/// ```rust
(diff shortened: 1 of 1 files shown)


jdm:
Could you split this into two separate commits? One for formatting changes, and the other for the actual implementation changes.

Vurich:
@jdm Sorry, I didn't intend to commit the formatting. I run `rustfmt` a lot when writing code because I'm lazy and can't be bothered to manually indent. I'll hard push a new commit without the formatting.

PR #59, opened 2017-08-17
Add retain function

Xaeroxe:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/59)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -605,6 +605,24 @@ impl<A: Array> SmallVec<A> {
             }
         }
     }
+
+    /// Retains only the elements specified by the predicate.
+    ///
+    /// In other words, remove all elements `e` such that `f(&e)` returns `false`.
+    /// This method operates in place and preserves the order of the retained
+    /// elements.
+    pub fn retain<F: FnMut(&A::Item) -> bool>(&mut self, mut f: F) {
+        let mut del = 0;
+        let len = self.len;
+        for i in 0..len {
+            if !f(&self[i]) {
+                del += 1;
+            } else if del > 0 {
+                self.swap(i - del, i);
+            }
+        }
+        self.truncate(len - del);
+    }
 }
 
 impl<A: Array> SmallVec<A> where A::Item: Copy {
@@ -1059,6 +1077,10 @@ pub mod tests {
     use std::borrow::ToOwned;
     #[cfg(not(feature = "std"))]
     use alloc::borrow::ToOwned;
+    #[cfg(feature = "std")]
+    use std::rc::Rc;
+    #[cfg(not(feature = "std"))]
+    use alloc::rc::Rc;
     #[cfg(not(feature = "std"))]
     use alloc::boxed::Box;
     #[cfg(not(feature = "std"))]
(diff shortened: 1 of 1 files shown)


lib.rs
```diff
@@ -605,6 +605,36 @@ impl<A: Array> SmallVec<A> {
...
+                            swap(&mut array.ptr().offset((i - del) as isize), &mut array.ptr().offset(i as isize));
+                        }
+                    }
+                }
+                if del > 0 {
+                    self.len = len - del;
+                }
+            }
+            Heap { ptr, capacity } => unsafe {
+                Vec::from_raw_parts(ptr, len, capacity).retain(f);
```
emilio: Wouldn't this drop the vector afterwards, causing a use after free the next time it's accessed? Also, I see nothing updating the `ptr` field, which may change I suppose.

lib.rs
```diff
@@ -605,6 +605,36 @@ impl<A: Array> SmallVec<A> {
...
+        match self.data {
+            Inline { ref mut array } => {
+                use std::mem::swap;
+                let mut del = 0;
+                unsafe {
+                    for i in 0..len {
+                        if !f(&*array.ptr().offset(i as isize)) {
+                            del += 1;
+                        } else if del > 0 {
+                            swap(&mut array.ptr().offset((i - del) as isize), &mut array.ptr().offset(i as isize));
```
emilio: If the item has a destructor, what calls it?

lib.rs
torkleyy: Yes, it would be dropped afterwards. I think you should get the new pointer, len and capacity from the vec and `mem::forget` it afterwards.

Xaeroxe:
@emilio I've addressed your comments and added tests, thanks for the help!

Xaeroxe:
Travis only failed on nightly.  Probably a regression in nightly.

lib.rs
```diff
@@ -605,6 +605,43 @@ impl<A: Array> SmallVec<A> {
...
+                        if !f(&*array.ptr().offset(i as isize)) {
+                            read(array.ptr().offset(i as isize));
+                            del += 1;
+                        } else if del > 0 {
+                            swap((array.ptr().offset((i - del) as isize) as *mut  A::Item).as_mut().unwrap(), (array.ptr().offset(i as isize) as *mut  A::Item).as_mut().unwrap());
+                        }
+                    }
+                }
+                if del > 0 {
+                    self.len = len - del;
```
emilio: nit: This can just be `self.len -= del;`, since `del` cannot be less than zero.

Xaeroxe:
Squished and addressed your nit.

lib.rs
```diff
@@ -605,6 +605,42 @@ impl<A: Array> SmallVec<A> {
...
+                    for i in 0..self.len {
+                        if !f(&*array.ptr().offset(i as isize)) {
+                            read(array.ptr().offset(i as isize));
+                            del += 1;
+                        } else if del > 0 {
+                            swap((array.ptr().offset((i - del) as isize) as *mut  A::Item).as_mut().unwrap(), (array.ptr().offset(i as isize) as *mut  A::Item).as_mut().unwrap());
+                        }
+                    }
+                }
+                if del > 0 {
```
emilio: nit: no need for the `if`, that's what I wanted to avoid in the first place :)

lib.rs
```diff
@@ -605,6 +605,42 @@ impl<A: Array> SmallVec<A> {
...
+            Inline { ref mut array } => {
+                use std::ptr::read;
+                let mut del = 0;
+                unsafe {
+                    for i in 0..self.len {
+                        if !f(&*array.ptr().offset(i as isize)) {
+                            read(array.ptr().offset(i as isize));
+                            del += 1;
+                        } else if del > 0 {
+                            swap((array.ptr().offset((i - del) as isize) as *mut  A::Item).as_mut().unwrap(), (array.ptr().offset(i as isize) as *mut  A::Item).as_mut().unwrap());
```
emilio: nit: Can't we just do `self.swap(i - del, i)`? We implement `DerefMut<[A::Item]>`.

emilio: Indeed if we do that and avoid the `ptr::read`, calling `truncate()` at the end, we can write this just with safe code, right?

Xaeroxe:
@emilio Thanks for the review!  The code has been greatly simplified and is no longer unsafe.

jdm:
@bors-servo: r=emilio,jdm
This is much easier to reason about now; thanks!

Xaeroxe:
More out of curiosity than any kind of dire need, does this project have any kind of regular release cycle?

jdm:
It is released whenever someone makes a PR that updated the Cargo.toml version.

mbrubeck:
This didn't occur to me until just now, but we might want to change the `retain` signature to pass `&mut A::Item` to the callback instead of `&A::Item`.  The Rust library team wishes they could make `Vec::retain` pass `&mut T`, and has used `&mut` for newer methods like `HashMap::retain`:

https://github.com/rust-lang/rust/pull/39560#issuecomment-279733693

Xaeroxe:
Why would you want to mutate in a retain?  It seems like you'd be better off doing
```Rust
for thing in things {
     // Mutate thing
}
things.retain(/*closure*/);
```

I like this approach better as it's more clear that there is a side effect happening.

mbrubeck:
> ```rust
> for thing in things {
>      // Mutate thing
> }
> things.retain(/*closure*/);
> ```
> 
> I like this approach better as it's more clear that there is a side effect happening.

Putting the mutation into a separate loop might require allocating additional memory to save intermediate state, while just using `retain` can use the results immediately.  For example, if you want to send a message to a bunch of MPSC channels, and discard any senders whose receivers have hung up:

`channels.retain(|sender| sender.send(message).is_ok());`

This can't be done in two separate loops like the code above, without saving additional state. 

Even in cases that can be easily separated, iterating over the collection twice instead can be a significant performance penalty.  And since the `retain` function does have unique (`&mut T`) access to the collection, there's no reason for it to give out only shared (`&T`) references.

Xaeroxe:
Alright that all makes sense.  The current retain got released in 0.4.3 which still has 0 downloads so if we hurry and yank it we can still fix this.

mbrubeck:
It might be less hassle to just fix this next time we make a release with breaking changes.

PR #61, opened 2017-08-22
[breaking change] Update retain to use &mut instead of &

Xaeroxe:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/61)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -611,11 +611,11 @@ impl<A: Array> SmallVec<A> {
     /// In other words, remove all elements `e` such that `f(&e)` returns `false`.
     /// This method operates in place and preserves the order of the retained
     /// elements.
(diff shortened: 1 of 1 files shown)


Xaeroxe:
cc @mbrubeck

mbrubeck:
Since this is a breaking change, we might want to consider if there are any other breaking changes we'd like to make, so that we can release them all together as smallvec 1.0.0.

Xaeroxe:
@mbrubeck well it's been almost 3 months and we haven't gotten any response on that.  Do we just want to go ahead and merge it?

mbrubeck:
@bors-servo r+

Yeah, sorry. I should have included this in the recent 0.5.0 release, but I forgot.  It can go in the next version.

PR #72, opened 2017-11-29
Add support for dedup, dedup_by, and dedup_by_key

hoggetaylor:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/72)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -91,36 +91,12 @@ pub trait VecLike<T>:
 
     /// Append an element to the vector.
     fn push(&mut self, value: T);
-    /// Removes consecutive repeated elements.
-    fn dedup(&mut self) where T: PartialEq<T>;
-    /// Removes consecutive repeated elements with a given equality relation.
-    fn dedup_by<F>(&mut self, same_bucket: F) where F: FnMut(&mut T, &mut T) -> bool;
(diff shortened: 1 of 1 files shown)


hoggetaylor:
This pull request adds support for dedup, dedup_by, and dedup_by_key as requested in #2

lib.rs
```diff
@@ -91,13 +91,37 @@ pub trait VecLike<T>:
 
     /// Append an element to the vector.
     fn push(&mut self, value: T);
+    /// Removes consecutive repeated elements.
+    fn dedup(&mut self) where T: PartialEq<T>;
```
mbrubeck: I'm not sure we want to add these as trait methods.  Currently this trait is very minimal; adding these methods would be a breaking change and a bit of a burden for other types that implement it.  (The use case for this trait is questionable anyway, and I think we might deprecate and remove it instead.)

If we did add these methods, then we should probably also add `truncate`, and then provide default implementations of the `dedup` methods based on that.

hoggetaylor: I have added a commit to remove these methods from the VecLike trait.

mbrubeck:
@bors-servo r+

Thanks!

PR #88, opened 2018-03-20
Add Array impl for [T; 0]

Vurich:
This seems useless but has benefits in generic contexts (plus, with smart-enough compiler optimisations this could conceivably compile to equivalent code to `Vec<T>`).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/88)


Vurich:
Wow, that was fast!

PR #89, opened 2018-04-01
Make SmallVec::retain compatible with Vec::remain

aloucks:
Replaces `&mut` with `&` in the function predicate so that it's compatible with `Vec::retain`:

```rust
pub fn retain<F>(&mut self, f: F) 
where
    F: FnMut(&T) -> bool
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/89)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -634,11 +634,11 @@ impl<A: Array> SmallVec<A> {
     /// In other words, remove all elements `e` such that `f(&e)` returns `false`.
     /// This method operates in place and preserves the order of the retained
     /// elements.
-    pub fn retain<F: FnMut(&mut A::Item) -> bool>(&mut self, mut f: F) {
+    pub fn retain<F: FnMut(&A::Item) -> bool>(&mut self, mut f: F) {
         let mut del = 0;
         let len = self.len;
         for i in 0..len {
-            if !f(&mut self[i]) {
+            if !f(&self[i]) {
                 del += 1;
             } else if del > 0 {
                 self.swap(i - del, i);
@@ -1710,7 +1710,7 @@ mod tests {
     fn test_retain() {
         // Test inline data storate
         let mut sv: SmallVec<[i32; 5]> = SmallVec::from_slice(&[1, 2, 3, 3, 4]);
-        sv.retain(|&mut i| i != 3);
+        sv.retain(|&i| i != 3);
         assert_eq!(sv.pop(), Some(4));
         assert_eq!(sv.pop(), Some(2));
         assert_eq!(sv.pop(), Some(1));
@@ -1718,7 +1718,7 @@ mod tests {
 
         // Test spilled data storage
         let mut sv: SmallVec<[i32; 3]> = SmallVec::from_slice(&[1, 2, 3, 3, 4]);
-        sv.retain(|&mut i| i != 3);
+        sv.retain(|&i| i != 3);
         assert_eq!(sv.pop(), Some(4));
         assert_eq!(sv.pop(), Some(2));
         assert_eq!(sv.pop(), Some(1));


tomaka:
If I remember correctly, the designers of `Vec::retain` mentioned that it was a mistake on their side to use `&` and not `&mut` (and they cannot fix it because of backward compatibility). The `&mut` solution seems strictly superior to me.

aloucks:
@tomaka I don't necessarily disagree, but I ran into an issue where trying to write something generic over Vec and SmallVec and got hung up by the differences in mutability.

emilio:
mut seems better to me as well, Why does the generic code don't work? Could you put an example?

aloucks:
I want to create blanket impl's over newtypes where the inner component is either a `Vec` or a `SmallVec`.  I was under the impression that SmallVec's interface should be compatible with Vec, but I agree that allowing the closure to take the item by `&mut` is a little more flexible. I'm able to work around it by copying the retain implementation, so it's not overly critical. Here's an example:

```rust
extern crate smallvec;

use smallvec::{SmallVec, Array};
use std::ops::{Deref, DerefMut, Rem};

macro_rules! tuple_deref {
    ($Type:ty, $Target:ty) => {
        impl Deref for $Type {
            type Target = $Target;
            fn deref(&self) -> &Self::Target {
                &self.0
            }
        }

        impl DerefMut for $Type {
            fn deref_mut(&mut self) -> &mut Self::Target {
                &mut self.0
            }
        }
    }
}

pub struct A(Vec<u32>);
pub struct B(SmallVec<[u32; 1]>);

tuple_deref!(A, Vec<u32>);
tuple_deref!(B, SmallVec<[u32; 1]>);

pub trait VecLike {
    type Item;
    fn is_empty(&self) -> bool;
    fn retain<F>(&mut self, f: F) where F: FnMut(&Self::Item) -> bool, Self: Sized;
}

impl<T> VecLike for Vec<T> {
    type Item = T;
    fn is_empty(&self) -> bool {
        Vec::<T>::is_empty(self)
    }
    fn retain<F>(&mut self, f: F) where F: FnMut(&Self::Item) -> bool, Self: Sized {
        Vec::<T>::retain(self, f)
    }
}

impl<T: Array> VecLike for SmallVec<T> {
    type Item = T::Item;
    fn is_empty(&self) -> bool {
        SmallVec::<T>::is_empty(self)
    }
    fn retain<F>(&mut self, mut f: F) where F: FnMut(&Self::Item) -> bool, Self: Sized {
        // SmallVec::<T>::retain(self, f)
        // https://docs.rs/smallvec/0.6.0/src/smallvec/lib.rs.html#637-648
        let mut del = 0;
        let len = self.len();
        for i in 0..len {
            if !f(&self[i]) {
                del += 1;
            } else if del > 0 {
                self.swap(i - del, i);
            }
        }
        self.truncate(len - del);
    }
}

pub trait Foobar<I> {
    fn foobar(&mut self);
}

impl<T, V, I> Foobar<I> for T 
    where V: VecLike<Item=I>,
          T: DerefMut<Target=V>,
          I: Copy + PartialEq<u32> + Rem<u32, Output=I>
{
    fn foobar(&mut self) {
        self.retain(|item| *item % 2 == 0);
    }
}

fn main() {
    let v = vec![1, 2, 3, 4];
    let mut a = A(v.clone());
    let mut b = B(SmallVec::<[u32; 1]>::from_vec(v.clone()));

    a.foobar();
    b.foobar();

    assert_eq!(a.as_slice(), &[2, 4]);
    assert_eq!(b.as_slice(), &[2, 4]);
}
```

emilio:
@aloucks I think you can just do:

```rust
fn retain<F>(&mut self, mut f: F) where F: FnMut(&Self::Item) -> bool, Self: Sized {
    SmallVec::<T>::retain(self, |e| f(&*e))
}
```

?

aloucks:
@emilio Thanks, that's much cleaner. I thought I had tried that, but apparently I had not!

PR #92, opened 2018-04-17
[WIP] Use a union to reduce the size of SmallVec

Amanieu:
Uses a union to eliminate the space used by the `enum` determinant. The new layout looks like this:
```rust
struct SmallVec<A> {
    capacity: usize,
    union {
        inline: ManuallyDrop<A>,
        heap: struct {
            ptr: *mut A::Item, 
            len: usize,
        } 
    }
}
```

The `capacity` field is used to determine which of the two union variants is active:
- If `capacity <= A::size()` then the inline variant is used and `capacity` holds the current length of the vector (number of elements actually in use).
- If `capacity > A::size()` then the heap variant is used and `capacity` holds the size of the memory allocation.

Since unions with drop are still currently unstable, this is kept behind a "union" cargo feature, which falls back to an implementation using an enum if disabled.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/92)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -17,6 +17,7 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(not(feature = "std"), feature(alloc))]
+#![cfg_attr(feature = "union", feature(untagged_unions))]
 #![deny(missing_docs)]
 
 #[cfg(not(feature = "std"))]
@@ -48,11 +49,11 @@ use std::iter::{repeat, FromIterator, IntoIterator};
 #[cfg(feature = "serde")]
 use std::marker::PhantomData;
 use std::mem;
+#[cfg(not(feature = "union"))]
+use std::mem::ManuallyDrop;
 use std::ops;
 use std::ptr;
 use std::slice;
-
-use SmallVecData::{Heap, Inline};
 
 /// Creates a [`SmallVec`] containing the arguments.
 ///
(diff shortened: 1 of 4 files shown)


Amanieu:
Hmm there seems to be a regression in the benchmark results. I will look into this.
```
 name                     old ns/iter  new ns/iter  diff ns/iter   diff %  speedup 
 bench_extend             188          328                   140   74.47%   x 0.57 
 bench_extend_from_slice  48           50                      2    4.17%   x 0.96 
 bench_from_slice         47           49                      2    4.26%   x 0.96 
 bench_insert             1,125        1,335                 210   18.67%   x 0.84 
 bench_insert_from_slice  94           104                    10   10.64%   x 0.90 
 bench_insert_many        402          385                   -17   -4.23%   x 1.04 
 bench_macro_from_elem    194          257                    63   32.47%   x 0.75 
 bench_macro_from_list    43           47                      4    9.30%   x 0.91 
 bench_push               382          351                   -31   -8.12%   x 1.09 
 bench_pushpop            282          671                   389  137.94%   x 0.42 
```

lib.rs
```diff
@@ -231,43 +232,90 @@ impl<'a, T: 'a> Drop for Drain<'a, T> {
...
+    }
+    #[inline]
+    fn from_inline(inline: A) -> SmallVecData<A> {
+        SmallVecData::Inline(ManuallyDrop::new(inline))
+    }
+    #[inline]
+    unsafe fn heap(&self) -> (*mut A::Item, usize) {
+        match *self {
+            SmallVecData::Heap(data) => data,
+            _ => unreachable!(),
```
emilio: debug_unreachable may be worth in these.

Vurich:
You might want to recommit this without rustfmt. Ideally, the rustfmt would be a seperate PR

Amanieu:
I'm going to close this since I haven't had time to work on it lately.

PR #93, opened 2018-05-07
Optimize `extend` and `from_elem`

Vurich:
This increases the amount of unsafe code, and so should be carefully checked. However, the use of unsafe code is quite benign and depends only on the invariant that `grow` always succeeds (which is depended on other places in the codebase)

Benchcmp results:

```
 name                     preopt.bench ns/iter  postopt.bench ns/iter  diff ns/iter   diff %  speedup 
 bench_extend             277                   89                             -188  -67.87%   x 3.11  
 bench_macro_from_elem    214                   61                             -153  -71.50%   x 3.51 
```

I took a crack at this after noticing how awful the generated assembly for `SmallVec::from_elem` is even for really simple types like integers. It's still not great but it at least now for the case of `Copy` types it gets autovectorised. Probably someone could make it even better if they spent more time on it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/93)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -810,7 +810,7 @@ impl<A: Array> SmallVec<A> where A::Item: Clone {
     /// ```
     pub fn from_elem(elem: A::Item, n: usize) -> Self {
         if n > A::size() {
-            ::std::vec::from_elem(elem, n).into()
+            vec![elem; n].into()
         } else {
             unsafe {
                 let mut arr: A = ::std::mem::uninitialized();
--- a/lib.rs
+++ b/lib.rs
@@ -21,7 +21,7 @@
 
 
 #[cfg(not(feature = "std"))]
-#[cfg_attr(test, macro_use)]
+#[macro_use]
 extern crate alloc;
 
 #[cfg(not(feature = "std"))]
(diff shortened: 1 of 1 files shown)


Vurich:
A comparison of the generated assembly for a simple (but probably common) case is [here](https://gist.github.com/Vurich/2d180d83821fc4f129b176a9a0f1dc63)

Vurich:
The reasoning behind using `std::vec::from_elem`/`vec!` over `.extend(iter::repeat(..).take(..))` is that `u8` and integer `0` have a fast path in `vec::from_elem` using `ptr::write_bytes` and `calloc`, respectively. We could conceivably also use `ptr::write_bytes` on `u8` and `0` when the vec would be inline but that would require a nightly feature (specialisation) and would only be noticeable on _huge_ inline arrays.

mbrubeck:
@bors-servo r+

Thanks!

mbrubeck:
If you have a project that needs this, let us know and we can publish a release right away.  (If not, it'll get published sooner or later, whenever the next release happens.)

PR #94, opened 2018-05-10
Use a union to reduce the size of SmallVec [v2]

arthurprs:
Building on top of #92 by @Amanieu 

I introduced `triple()` and `triple_mut()` which removed almost all of the runtime overhead. Performance is very comparable.

```
 name                     master:: ns/iter  union:: ns/iter  diff ns/iter   diff %  speedup 
 bench_extend             45                47                          2    4.44%   x 0.96 
 bench_extend_from_slice  45                43                         -2   -4.44%   x 1.05 
 bench_from_slice         45                44                         -1   -2.22%   x 1.02 
 bench_insert             615               622                         7    1.14%   x 0.99 
 bench_insert_from_slice  101               99                         -2   -1.98%   x 1.02 
 bench_insert_many        309               266                       -43  -13.92%   x 1.16 
 bench_macro_from_elem    41                37                         -4   -9.76%   x 1.11 
 bench_macro_from_list    40                42                          2    5.00%   x 0.95 
 bench_push               381               370                       -11   -2.89%   x 1.03 
 bench_pushpop            404               420                        16    3.96%   x 0.96 
 bench_remove             458               436                       -22   -4.80%   x 1.05 
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/94)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ documentation = "http://doc.servo.org/smallvec/"
 
 [features]
 std = []
+union = []
 default = ["std"]
 
 [lib]
@@ -19,6 +20,8 @@ name = "smallvec"
 path = "lib.rs"
 
 [dependencies]
+unreachable = "1.0.0"
+debug_unreachable = "0.1"
 serde = { version = "1", optional = true }
 
 [dev_dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,6 @@ path = "lib.rs"
 
 [dependencies]
 unreachable = "1.0.0"
-debug_unreachable = "0.1"
 serde = { version = "1", optional = true }
 
 [dev_dependencies]
--- a/lib.rs
+++ b/lib.rs
@@ -17,6 +17,7 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(not(feature = "std"), feature(alloc))]
+#![cfg_attr(feature = "union", feature(untagged_unions))]
 #![deny(missing_docs)]
 
 
@@ -30,6 +31,13 @@ use alloc::Vec;
 #[cfg(feature = "serde")]
 extern crate serde;
 
+extern crate unreachable;
+use unreachable::UncheckedOptionExt;
+
+#[cfg(not(feature = "union"))]
+#[macro_use]
+extern crate debug_unreachable;
+
 #[cfg(not(feature = "std"))]
 mod std {
     pub use core::*;
@@ -41,6 +49,8 @@ use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::iter::{IntoIterator, FromIterator, repeat};
 use std::mem;
+#[cfg(not(feature = "union"))]
+use std::mem::ManuallyDrop;
 use std::ops;
 use std::ptr;
 use std::slice;
@@ -52,8 +62,6 @@ use serde::ser::{Serialize, Serializer, SerializeSeq};
 use serde::de::{Deserialize, Deserializer, SeqAccess, Visitor};
 #[cfg(feature = "serde")]
 use std::marker::PhantomData;
-
-use SmallVecData::{Inline, Heap};
 
 /// Creates a [`SmallVec`] containing the arguments.
 ///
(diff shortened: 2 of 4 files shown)


Vurich:
Your benchmark changes will definitely cause conflicts with #97, I don't know if you want to add your bench changes as a seperate PR to avoid that (I don't mind rolling them into #97 instead if you like).

arthurprs:
@Vurich fell free to put them into #97 and I'll rebase this.

arthurprs:
I rebased on master now, the only change I kept was to add the return type to remove_no_inline (to prevent over optimization, specially for Vec).

lib.rs
```diff
@@ -14,6 +14,17 @@
...
+//!
+//! ## `union` feature
+//!
+//! When the `union` feature is enabled `smallvec` will track its state (inline or spilled)
+//! without the use of an enum tag, reducing the size of the `smallvec` by one machine word.
+//! This means that there is potentially no space overhead compared to `Vec`.
+//! Note that `smallvec` can still be larger than `Vec` if the inline buffer is larger than two
+//! machine words.
+//!
+//! To use this feature add `features = ["uuid"]` in the `smallvec` section of Cargo.toml.
```
psii: Should be `union` instead of `uuid`, right?

arthurprs: Ops, will fix.

arthurprs:
It'd be nice to get this moving. It may be interesting for rustc as well.

jdm:
@mbrubeck Do you have time to review these changes?

lib.rs
```diff
@@ -350,14 +419,25 @@ impl<A: Array> SmallVec<A> {
     /// ```
     #[inline]
     pub fn from_vec(mut vec: Vec<A::Item>) -> SmallVec<A> {
-        let (ptr, cap, len) = (vec.as_mut_ptr(), vec.capacity(), vec.len());
-        mem::forget(vec);
+        if vec.capacity() <= A::size() {
```
mbrubeck: The documentation says that `from_vec` does not copy the elements.  I think we should keep the current behavior of keeping the items on the heap.  Users who want to move them inline if possible can use `shrink_to_fit` or other methods.

lib.rs
```diff
@@ -301,7 +370,7 @@ impl<A: Array> Drop for SmallVecData<A> {
 /// assert!(v.spilled());
 /// ```
 pub struct SmallVec<A: Array> {
-    len: usize,
+    capacity: usize,
```
mbrubeck: Using `capacity` here is a little confusing to me, but I'm not sure what a better name would be.  Maybe something like `len_cap`?  Maybe `capacity` is okay, but there should be a comment here describing the semantics.

lib.rs
```diff
@@ -399,41 +480,59 @@ impl<A: Array> SmallVec<A> {
...
     #[inline]
     pub fn capacity(&self) -> usize {
-        match self.data {
-            Inline { .. } => A::size(),
-            Heap { capacity, .. } => capacity,
+        self.triple().2
+    }
+
+    #[inline]
+    fn triple(&self) -> (*const A::Item, usize, usize) {
```
mbrubeck: Please add a comment to say what the return value means.

lib.rs
```diff
@@ -556,12 +661,12 @@ impl<A: Array> SmallVec<A> {
...
-            unsafe {
-                let last_index = self.len - 1;
-                self.set_len(last_index);
-                ptr::read(end_ptr.offset(last_index as isize));
+        unsafe {
+            let (ptr, len_ptr, _) = self.triple_mut();
+            while len < *len_ptr {
+                let last_index = *len_ptr - 1;
+                *len_ptr = last_index;
+                ptr::read(ptr.offset(last_index as isize));
```
mbrubeck: This could use `ptr::drop_in_place` instead of `ptr::read`.

lib.rs
```diff
@@ -1066,12 +1163,15 @@ impl<A: Array> Default for SmallVec<A> {
...
-            let ptr = self.as_ptr();
-            for i in 0 .. self.len {
-                ptr::read(ptr.offset(i as isize));
+            if self.spilled() {
+                let (ptr, len) = self.data.heap();
+                Vec::from_raw_parts(ptr, len, self.capacity);
+            } else {
+                let ptr = self.as_ptr();
+                for i in 0..self.len() {
+                    ptr::read(ptr.offset(i as isize));
```
mbrubeck: This could use `ptr::drop_in_place` instead.

lib.rs
Amanieu: This is unfortunately not possible with the new design: there is no way of representing a heap-allocated buffer with a capacity less than `A::size()`. If `self.capacity <= A::size()` then the code assumes that the data is in the inline buffer.

mbrubeck: Oh, right.  In that case the documentation should be updated.

Cargo.toml
```diff
@@ -12,13 +12,16 @@ documentation = "http://doc.servo.org/smallvec/"
...
+union = []
 default = ["std"]
 
 [lib]
 name = "smallvec"
 path = "lib.rs"
 
 [dependencies]
+unreachable = "1.0.0"
+debug_unreachable = "0.1"
```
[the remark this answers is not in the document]
arthurprs: Makes sense, I'll remove the dependency and add the equivalent macro.

lib.rs
arthurprs: Good point.

mbrubeck:
Great, thanks!  I think this is ready but I just want to double-check all the unsafe code before merging it (and give a chance for others to do the same).

PR #103, opened 2018-07-18
Panic-safety fixes

### LINKED ISSUE #96: `SmallVec::insert_many` is unsound
[Gist here](https://gist.github.com/Vurich/ffa55d8e69377a64c9708e07c2db2a1b). A solution to this would be to set `len = index` before iterating. Obviously this would cause leaks but we're already leaking data.

(Sorry for the very verbose explanation in the linked gist but I wanted to link it as part of an article whose audience might not immediately understand the problem).

### LINKED ISSUE #101: from_elem is unsound when clone can panic
If `clone` panics during `SmallVec::from_elem` then uninitialized memory is dropped. [Original report by dbaupp on Reddit.]

This bug was introduced by #93 which is not yet included in the latest published release (0.6.2).

The suggested solution is to use something like `SetLenOnDrop` to make sure the length is set correctly when destructors run, without inhibiting optimizations.

[1]: https://www.reddit.com/r/rust/comments/8l8l7k/improving_smallvecs_speed_by_60_and_why_that/dzdt4x2/
[2]: https://github.com/rust-lang/rust/blob/5205ae8bdc2991eecf3bfbb58ed8f56c0673e738/src/liballoc/vec.rs#L1489-L1516

mbrubeck:
* Make from_elem panic-safe.  Fixes #101.
* Make insert_many panic-safe.  Fixes #96.

r? @SimonSapin or @jdm.  cc @Vurich

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/103)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -771,24 +771,33 @@ impl<A: Array> SmallVec<A> {
         unsafe {
             let old_len = self.len();
             assert!(index <= old_len);
-            let ptr = self.as_mut_ptr().offset(index as isize);
+            let mut ptr = self.as_mut_ptr().offset(index as isize);
+
+            // Move the trailing elements.
             ptr::copy(ptr, ptr.offset(lower_size_bound as isize), old_len - index);
-            for (off, element) in iter.enumerate() {
-                if off < lower_size_bound {
-                    ptr::write(ptr.offset(off as isize), element);
-                    let len = self.len() + 1;
-                    self.set_len(len);
-                } else {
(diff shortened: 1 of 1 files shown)


mbrubeck:
Note: This PR also speeds up our `insert_many` benchmark by about 40%, and does not change the performance of our `from_elem` benchmarks.

jdm:
@bors-servo r+
I like the from_elem solution in particular.

PR #95, opened 2018-05-17
Add benchmarks for Vec to compare to SmallVec

Vurich:
Preliminary results aren't good, although no benches check for cache efficiency (for example `Vec<Vec<T>>` vs `Vec<SmallVec<[T; N]>>`).

Here's a comparison. I used `sed` to split the `_vec` benchmarks out and rename them to have the same name, so `cargo benchcmp` can understand them.

```
 name                     vec.bench ns/iter          smallvec.bench ns/iter          diff ns/iter   diff %  speedup 
 bench_extend             69                         53                                       -16  -23.19%   x 1.30 
 bench_extend_from_slice  64                         52                                       -12  -18.75%   x 1.23 
 bench_from_slice         34                         52                                        18   52.94%   x 0.65 
 bench_insert             1,228                      1,202                                    -26   -2.12%   x 1.02 
 bench_macro_from_elem    50                         66                                        16   32.00%   x 0.76 
 bench_macro_from_list    33                         47                                        14   42.42%   x 0.70 
 bench_push               359                        424                                       65   18.11%   x 0.85 
 bench_pushpop            251                        348                                       97   38.65%   x 0.72 
```

Closes #87

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/95)

PR #105, opened 2018-07-19
Version 0.6.3

mbrubeck:
Changes in this release:

* Fix possible double-free in `insert_many` when passed an iterator that panics in `next` (#103)
* Add a new `union` feature (disabled by default) that reduces the size of the SmallVec struct (#94)
* Improve performance of `extend` and `from_elem` (#93)
* Improve performance of `drop` (#100)
* Update to build without `std` feature on current Rust nightly (#104)
* Additional benchmarks (#95, #97)
* Update dev-dependency on `bincode` (#102)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/105)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "smallvec"
-version = "0.6.1"
+version = "0.6.3"
 authors = ["Simon Sapin <simon.sapin@exyr.org>"]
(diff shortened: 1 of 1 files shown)


KiChjang:
A minor detail here, but why did we skip 0.6.2?

PR #107, opened 2018-07-27
Allow smallvec! with non-Copy items

### LINKED ISSUE #98: smallvec! macro requires T to be Copy
The `smallvec!` macro, when using the `smallvec![a,b,c]` form, requires that `T: Copy`. 

This is because the macro expands to use `SmallVec::from_slice`. 

```
    ($($x:expr),*) => ({
        SmallVec::from_slice(&[$($x),*])
    });
```

Perhaps it could instead expand into something like...

```
let s = SmallVec::new();
s.push(a);
s.push(b);
s.push(c);
s
```
... and remove the need for `T: Copy`?

RReverser:
When count of items is smaller or equal than the target inline size, macro will use `SmallVec::push`, but when it's large enough, it passes data on to `vec!` macro for in-place heap allocation and then uses `SmallVec::from_vec`.

This trick gives ~3.5x performance increase on `bench_macro_from_list` compared to `SmallVec::with_capacity`.

Fixes #98.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/107)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -113,11 +113,20 @@ use std::marker::PhantomData;
 
 #[macro_export]
 macro_rules! smallvec {
+    // count helper: transform any expression into 1
+    (@one $x:expr) => (1usize);
     ($elem:expr; $n:expr) => ({
-        SmallVec::from_elem($elem, $n)
+        $crate::SmallVec::from_elem($elem, $n)
     });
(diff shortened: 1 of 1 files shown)


mbrubeck:
@bors-servo r+

Thanks!

PR #109, opened 2018-08-05
Improved debug implementation.

thelearnerofcode:
Before SmallVec would use "{:?}" style debug info no matter what the programmer specified. This pull request makes it much more inline with what Vec does by replacing the previous implementation with DebugList.

See: [https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_list](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_list)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/109)


mbrubeck:
@bors-servo r+

Thank you!

PR #114, opened 2018-08-07
Test in Rust 1.20 on Travis

mbrubeck:
This will ensure we don't increase the minimum supported Rust version without noticing it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/114)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: rust
 rust:
+  - 1.20.0
   - nightly
   - beta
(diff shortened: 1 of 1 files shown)


mbrubeck:
> I assume this is not a hard requirement provided we notice and do a breaking bump, right?

Right, this does not imply any particular guarantee of support in future versions.  It's just to help us track which versions are supported.

PR #117, opened 2018-08-20
new constructors: from_buf_and_len(_unchecked)

llogiq:
Those functions allow to create an inline SmallVec supplying both buf and len arguments, so only a part of the buffer is used. The unchecked variant doesn't check if the length is less or equal than the buffer length.

This allows users to set up a partially used in-band buffer in one step (instead of using `from_buf` and calling `truncate` or `set_len` later).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/117)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -480,6 +480,47 @@ impl<A: Array> SmallVec<A> {
         }
     }
 
+    /// Constructs a new `SmallVec` on the stack from an `A` without
+    /// copying elements. Also sets the length, which must be less or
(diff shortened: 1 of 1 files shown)


mbrubeck:
@bors-servo r+

Thanks!

llogiq:
Thanks @mbrubeck :+1:

llogiq:
By the way, @mbrubeck can we get a new release with this?

PR #118, opened 2018-08-22
bench and optimize `from_slice`

llogiq:
This adds "bench_from_slice(_vec)(_small)" benches, renaming the previous versions to "bench_from_iter*" (because they were using `From<&[T]>`, which calls `FromIterator` internally) and optimizes the `from_slice` constructor considerably.

Before:
```
test bench_from_slice                  ... bench:          42 ns/iter (+/- 0)
test bench_from_slice_small            ... bench:          12 ns/iter (+/- 0)
```
After:
```
test bench_from_slice                  ... bench:          27 ns/iter (+/- 0)
test bench_from_slice_small            ... bench:           8 ns/iter (+/- 0)
```


llogiq:
travis error is unrelated – current nightly fails to build master as well.

mbrubeck:
Looks like this needs a `#[cfg(not(feature = "std"))] use alloc::borrow::ToOwned;`.

llogiq:
I'll settle for `.to_vec()` instead.

llogiq:
@mbrubeck this is ready to merge now.

PR #122, opened 2018-08-23
fix from_slice possible overallocation

llogiq:
@arthurps found that the code would underreport the vec's capacity
if `slice.to_vec()` allocated more than `len` entries. This
changes the code to set the correct capacity as reported from the
`vec`.


mbrubeck:
cc #118

PR #123, opened 2018-08-27
First specialization

llogiq:
I added a new `specialization` feature that will specialize `From<&[_: Copy]>` to use `from_slice`, which offers a nice performance boost.

Alas, I could not get any measurable perf improvement on `insert_many` or `extend` yet, so I'll leave them out for now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/123)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -31,6 +31,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(not(feature = "std"), feature(alloc))]
 #![cfg_attr(feature = "union", feature(untagged_unions))]
+#![cfg_attr(feature = "specialization", feature(specialization))]
 #![deny(missing_docs)]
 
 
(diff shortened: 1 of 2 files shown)


lib.rs
```diff
@@ -1157,10 +1158,25 @@ where A::Item: Deserialize<'de>,
...
 impl<'a, A: Array> From<&'a [A::Item]> for SmallVec<A> where A::Item: Clone {
+    #[cfg(not(feature = "specialization"))]
     #[inline]
     fn from(slice: &'a [A::Item]) -> SmallVec<A> {
         slice.into_iter().cloned().collect()
     }
+
+    #[cfg(feature = "specialization")]
+    #[inline]
+    default fn from(slice: &'a [A::Item]) -> SmallVec<A> {
```
nox: Doesn't that mean that one can specialise the implementation from outside the crate? Shouldn't we be specialising an internal trait used by this impl?

llogiq: I'll add an internal SpecFrom trait then.

llogiq: Done.

llogiq:
This should now be ready for merging.

PR #130, opened 2018-11-13
Add from_raw_parts.

heycam:
I have a need to create a `SmallVec` using a heap allocation that I control.  Something that works just like `Vec::from_raw_parts` fits the bill.  The PR here adds a similar method, with the important difference that it panics if asked to conjure up a `SmallVec` that would use its inline storage rather than spilled, heap storage.

This is a dependency of my work in https://bugzilla.mozilla.org/show_bug.cgi?id=1474793.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/130)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -940,6 +940,86 @@ impl<A: Array> SmallVec<A> {
     {
         self.dedup_by(|a, b| key(a) == key(b));
     }
+
+    /// Creates a `SmallVec` directly from the raw components of another
+    /// `SmallVec`.
+    ///
+    /// # Safety
+    ///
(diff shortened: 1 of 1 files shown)


emilio:
Would it be better to unconditionally use `from_heap` in `SmallVec::from_vec`? It looks to me like it's pointless to move the data to the stack.

emilio:
Oh yeah I guess that wouldn't be safe, nvm

mbrubeck:
Published: https://crates.io/crates/smallvec/0.6.6

PR #133, opened 2018-11-29
Add a new feature, `may_dangle`.

### LINKED ISSUE #132: Add opt-in `may_dangle` support
A couple of times now I've tried to use `SmallVec` as a drop-in replacement for `Vec` within rustc, and it hasn't worked because `SmallVec` doesn't have `#[may_dangle]` annotations. (https://github.com/rust-lang/rust/pull/55525 is one case; https://github.com/rust-lang/rust/pull/56269 is the other.) That annotation is unstable so I think it would have to be behind a feature flag, but that would suffice for my purposes.

I'm not sure about the right way to do this, though. I tried changing this line:

>impl<A: Array> Drop for SmallVec<A> {

to this:

> unsafe impl<#[may_dangle] A: Array> Drop for SmallVec<A> {

For the current code I'm working on (https://github.com/rust-lang/rust/pull/56269), where `Vec` works but `SmallVec` doesn't, this reduced the number of borrow checker errors from 4 to 2. So it helped, but didn't solve things completely. Maybe there's a better way to do it.

nnethercote:
This commit adds a `may_dangle` annotation to `drop`, matching `Vec`.
This feature increases the ability of `SmallVec` to be used as a drop-in
replacement for `Vec`. A feature is necessary because `may_dangle` is
Nightly-only.

Fixes #132.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/133)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "smallvec"
-version = "0.6.6"
+version = "0.6.7"
 authors = ["Simon Sapin <simon.sapin@exyr.org>"]
 license = "MIT/Apache-2.0"
 repository = "https://github.com/servo/rust-smallvec"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ std = []
 union = []
 default = ["std"]
 specialization = []
+may_dangle = []
 
 [lib]
 name = "smallvec"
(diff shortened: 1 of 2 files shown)


mbrubeck:
Published: https://crates.io/crates/smallvec/0.6.7

PR #134, opened 2018-12-05
Add a push_light method and use likely/unlikely intrinsics

Zoxc:
This adds a `push_light` alternative to `push` which is optimized for inlining and for vectors which fit inside the fixed size.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/134)

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -9,10 +9,12 @@ use smallvec::{ExtendFromSlice, SmallVec};
 
 const VEC_SIZE: usize = 16;
 const SPILLED_SIZE: usize = 100;
+const LARGE_SIZE: usize = 10000;
 
 trait Vector<T>: for<'a> From<&'a [T]> + Extend<T> + ExtendFromSlice<T> {
     fn new() -> Self;
     fn push(&mut self, val: T);
+    fn push_light(&mut self, val: T);
     fn pop(&mut self) -> Option<T>;
     fn remove(&mut self, p: usize) -> T;
     fn insert(&mut self, n: usize, val: T);
@@ -28,6 +30,8 @@ impl<T: Copy> Vector<T> for Vec<T> {
     fn push(&mut self, val: T) {
         self.push(val)
     }
+
+    fn push_light(&mut self, val: T) {}
 
     fn pop(&mut self) -> Option<T> {
         self.pop()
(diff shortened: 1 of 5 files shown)


benches/bench.rs
```diff
@@ -59,6 +63,14 @@ impl<T: Copy> Vector<T> for SmallVec<[T; VEC_SIZE]> {
         self.push(val)
     }
 
+    #[cfg(feature = "push_light")]
+    fn push_light(&mut self, val: T) {
+        self.push_light(val)
+    }
+
+    #[cfg(not(feature = "push_light"))]
+    fn push_light(&mut self, _: T) {}
```
nox: I don't understand why this is defined if the feature is missing, and why it does nothing.

Zoxc: It's to ensure things still compile

nox: It compiles sure, but it doesn't do the right thing, why should it be a noop if the feature is missing? Sounds like a way to have silent bugs.

benches/bench.rs
```diff
@@ -9,10 +9,12 @@ use smallvec::{ExtendFromSlice, SmallVec};
 
 const VEC_SIZE: usize = 16;
 const SPILLED_SIZE: usize = 100;
+const LARGE_SIZE: usize = 10000;
 
 trait Vector<T>: for<'a> From<&'a [T]> + Extend<T> + ExtendFromSlice<T> {
     fn new() -> Self;
     fn push(&mut self, val: T);
+    fn push_light(&mut self, val: T);
```
nox: IMO this should either not exist when the feature is not enabled, or delegate to `push` in the default implementation.

benches/bench.rs
Zoxc: It's not meaningful to benchmark a feature if it's disabled

nox: Oh ok sorry I didn't notice this was an implementation for benchmarks.

Zoxc:
It probably does make sense to make it the default behavior.

Vurich:
Additionally, I would mark this function `#[inline]` and not `#[inline(always)]`. You basically never want `#[inline(always)]` unless you can produce benchmarks proving that it's slower without it. LLVM's inlining heuristics are really good and you can mess up cold code in other functions or people optimising for code size if you use `inline(always)`.

Vurich:
Was this ever made the default behaviour? It seems like an easy win.

mbrubeck:
We can't make this the default as written, because even without the `likely` feature it still depends on the unstable `assume` intrinsic. If this is still a win without any intrinsics then I'd be happy to replace the current `push` with `push_light` (and keep the intrinsics behind an optional feature).

mbrubeck:
I pushed a rebased version of this PR plus a commit to make `push_light` the default, in case anyone wants to experiment with it:

https://github.com/servo/rust-smallvec/compare/master...mbrubeck:light2

At a glance, it looks like this is still a win for `bench_push_small` even without the `assume` or `likely`/`unlikely` intrinsics.

Vurich:
You can always have a `cfg(feature = nightly)`

PR #135, opened 2018-12-19
Update union code to use ManuallyDrop

bluss:
At the time of this writing, this breaking change is *proposed*, see rust-lang/rust/pull/56440.
Smallvec would have to be updated before the change can be merged upstream. Thanks!

A Rust breaking change to the untagged_unions feature means that no
union fields may have destructors, so we need use ManuallyDrop also
around the inline union field.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/135)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -60,7 +60,6 @@ use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::iter::{IntoIterator, FromIterator, repeat};
 use std::mem;
-#[cfg(not(feature = "union"))]
 use std::mem::ManuallyDrop;
 use std::ops;
 use std::ptr;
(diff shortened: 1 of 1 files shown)


bluss:
This change takes us further on the road to stable unions for SmallVec

mbrubeck:
@bors-servo r+

Thanks!

bluss:
I've reviewed the surrounding code, and it's really hard to talk about soundness with the unsafe code guidelines making a rumbling noise in the background, but without clear pronouncements.

One place of worry about this PR could be if there is a practical difference of uninitialized value inside `ManuallyDrop` inside a union vs just the previous uninitialized value inside a union. Since this is nightly only code, we could “upgrade” to MaybeUninit.

With MaybeUninit the question remains about *partial* initialization and the soundness of forming `&A` references when the `A` value is partially initialized, and if there is a way to do partial initialization correctly with raw pointers.

Issue #147, opened 2019-06-04
error when updating rustc


ehuss:
I'm trying to update rust-lang/rust to the latest smallvec (0.6.7 ➙ 0.6.9), and an running into an issue where rustc fails to compile:

> error: internal compiler error: src/librustc/ty/relate.rs:773: impossible case reached: can't relate: '_#0r with Type(RustaceansAreAwesome)

I have narrowed down the change that breaks this to #137.

I don't know anything about smallvec and very little about rustc, so if anyone can help figure out what's wrong, I'd appreciate it.  I'll keep digging a little, but it is unlikely that I'll be able to do more.  This can be reproduced with `cargo update -p smallvec; ./x.py build` (on master, 021a5033098ff0e3f7126acc7ac35149d325f16d).

<details><summary>backtrace</summary>
<pre>
thread 'rustc' panicked at 'Box<Any>', src/librustc_errors/lib.rs:637:9
stack backtrace:
   0: std::sys_common::backtrace::print
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::default_hook
   3: rustc::util::common::panic_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::begin_panic
   6: rustc_errors::Handler::bug
   7: rustc::util::bug::opt_span_bug_fmt::{{closure}}
   8: rustc::ty::context::tls::with_opt::{{closure}}
   9: rustc::ty::context::tls::with_context_opt
  10: rustc::ty::context::tls::with_opt
  11: rustc::util::bug::opt_span_bug_fmt
  12: rustc::util::bug::bug_fmt
  13: <rustc::ty::subst::Kind as rustc::ty::relate::Relate>::relate
  14: <&mut I as core::iter::traits::iterator::Iterator>::next
  15: <smallvec::SmallVec<A> as core::iter::traits::collect::FromIterator<<A as smallvec::Array>::Item>>::from_iter
  16: <core::result::Result<T,E> as rustc::ty::context::InternIteratorElement<T,R>>::intern_with
  17: rustc::ty::relate::relate_substs
  18: <rustc::ty::sty::ExistentialTraitRef as rustc::ty::relate::Relate>::relate
  19: rustc::infer::higher_ranked::<impl rustc::infer::combine::CombineFields>::higher_ranked_sub
  20: rustc::infer::InferCtxt::commit_if_ok
  21: rustc::traits::select::SelectionContext::match_poly_trait_ref
[...]

PR #151, opened 2019-06-06
Fix using `grow` to the same size.

### LINKED ISSUE #148: use-after-free when growing to the same size
Attempting to call `grow` on a spilled SmallVec with a value equal to the current capacity causes it to free the existing data.

```rust
let mut v: SmallVec<[u8; 2]> = SmallVec::new();
v.push(0);
v.push(1);
// Cause it to spill.
v.push(2);
assert_eq!(v.capacity(), 4);
// grow with the same capacity
v.grow(4);
// The backing storage is now freed, here be dragons.
println!("{:?}", v);
```

In the [`grow` method](https://github.com/servo/rust-smallvec/blob/19de50108d403efaa7cd979eac3bb97a4432fd4b/lib.rs#L651-L667) it falls through to deallocate.

I believe something like the following is the fix:

```diff
diff --git a/lib.rs b/lib.rs
index 5af32ba..b81b35a 100644
--- a/lib.rs
+++ b/lib.rs
@@ -664,6 +664,8 @@ impl<A: Array> SmallVec<A> {
                 if unspilled {
                     return;
                 }
+            } else {
+                return;
             }
             deallocate(ptr, cap);
         }
```

ehuss:
Using `grow` on a spilled SmallVec to the current capacity would free the backing storage when it shouldn't.

Fixes #148

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/151)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -665,6 +665,8 @@ impl<A: Array> SmallVec<A> {
                 if unspilled {
                     return;
                 }
+            } else {
+                return;
             }
             deallocate(ptr, cap);
         }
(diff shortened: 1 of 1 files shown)


emilio:
@bors-servo r+

Do you need a release after your fixes?

Also, thanks for the patch, and sorry I misunderstood #147 initially, I thought that a rustc update was breaking when compiling `SmallVec`, not the other way around!

Thanks for all the fixes, very welcome :)

ehuss:
Yea, if you could make a new release, I'd appreciate it.

PR #155, opened 2019-06-15
New modifications

### LINKED ISSUE #154: Constant generic
Constant generic is becoming more mature and I would like to know if is someone porting the existing code to use it behind a feature flag. If not, I might do it myself if you guys don't mind it.

c410-f3r:
Fixes #154 and some other things.

- Adds the `alloc` feature and doesn't remove the `std` feature because of `std::io::Write`.
- Replaces `mem::uninitialized` with `MaybeUninit`.
- Uses `NonNull<T>` instead of `*mut T`.
- Removes deprecated `VecLike`.
- Uses  `core::hint::unreachable_unchecked` instead of `unchecked`.
- Splits bloated `lib.rs` file into smaller modules for better readability, localization and maintainability.
- Bumps minimum compiler version to `1.36`.
- Uses the new `2018` ~epoch~ version.
- Adds new scripts for testing in development and CI.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/155)


jdm:
Could you please redo these changes so there's a separate commit for each of the points you mentioned? This is really difficult to review in its current condition.

c410-f3r:
@jdm Sure

lib.rs
```diff
@@ -258,25 +258,25 @@ impl<A: Array> SmallVecData<A> {
     #[inline]
     unsafe fn inline(&self) -> &A {
         match *self {
-            SmallVecData::Inline(ref a) => a,
+            SmallVecData::Inline(ref a) => &*a.as_ptr(),
```
mbrubeck: I believe it's unsound to return an `&A` that points to a possibly-uninitialized `A`.  This method should be changed to return a raw pointer. (The callers of this method only need raw pointers anyways.)

lib.rs
```diff
@@ -258,25 +258,25 @@ impl<A: Array> SmallVecData<A> {
...
             _ => debug_unreachable!(),
         }
     }
     #[inline]
     unsafe fn inline_mut(&mut self) -> &mut A {
         match *self {
-            SmallVecData::Inline(ref mut a) => a,
+            SmallVecData::Inline(ref mut a) => &mut *a.as_mut_ptr(),
```
mbrubeck: Like `inline` above, this method should be changed to return a raw pointer.

c410-f3r:
Thanks @mbrubeck

c410-f3r:
`1.36` was released and Travis now checks all variants of Rust.

Can I squash commits?

jdm:
@SimonSapin How does this work relate to #157?

SimonSapin:
I haven’t looked at the diff, but based on the description a lot of it is similar or equivalent. Only with a different approach: incremental changes v.s. rewrite.

c410-f3r:
Based on this current work, I can apply @SimonSapin work in future PRs

c410-f3r:
Constant generics are now fully working. Is anything preventing this PR from being merged?

SimonSapin:
While experimenting with const generics in a branch is fine, I would prefer not to have them in the `master` branch or on a crates.io release of this crate, at this time.

> Constant generics are now fully working.

I realize that’s not exactly what you meant, but this is an overstatement.

Using `#![feature(const_generics)]` makes current nigthlies print ``warning: the feature `const_generics` is incomplete and may cause the compiler to crash``. As far as I know this kind of warning is unprecedented in Rust. `smallvec`’s tests might be passing with const generic now, but there is no telling whether it’ll regress in a future Nightly.

c410-f3r:
> I realize this is not exactly what you meant, but that’s an overstatement.

I am sorry for the misunderstanding. Early versions of this PR contained a non-functional implementation for constant generics and it is now properly fixed.

> As far as I know this kind of warning is unprecedented in Rust. smallvec’s tests might be passing with const generic now, but there is no telling whether it’ll regress in a future Nightly.

It is the own nature of unstable Rustc features to break existing code in future releases, the same reason can be applied to already included unstable features of `smallvec`.

The current code is structured in a way that **no** constant generic stuff is invoked without explicitly telling otherwise by setting the feature, therefore, everything on stable will work without problems.

Nevertheless, if you guys feel comfortable enough to merge this code without constant generics, I can extract it into a separate PR.

c410-f3r:
- Constant generics inside macros has been fixed in https://github.com/rust-lang/rust/pull/63083#issuecomment-516192572.
- The job for Rust nightly in CI is now allowed to fail.

emilio:
I think most of this stuff has landed already incrementally... Should this be closed?

emilio:
(Like, constant generics and other improvements are still welcome I guess, but this PR as is is pretty unlikely to land)

mbrubeck:
I cherry-picked and landed some of the changes from this that weren't covered by other PRs.  I'm closing this now but feel free to open new PRs for any remaining changes.  Smaller, incremental patches are appreciated if possible.

Thanks!

PR #161, opened 2019-08-20
Move use of 'default fn' to its own module

Centril:
This is the proper way to do `cfg(...)` gating when dealing with an unstable feature that introduces syntax.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/161)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -1253,12 +1253,7 @@ trait SpecFrom<A: Array, S> {
 }
 
 #[cfg(feature = "specialization")]
-impl<'a, A: Array> SpecFrom<A, &'a [A::Item]> for SmallVec<A> where A::Item: Clone {
-    #[inline]
-    default fn spec_from(slice: &'a [A::Item]) -> SmallVec<A> {
-        slice.into_iter().cloned().collect()
-    }
-}
+mod specialization;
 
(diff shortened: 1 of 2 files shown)


emilio:
Can you elaborate? Why is it the right way, or better than the current code?

This moves one of the specialized functions to its own module, but not the rest, so now you need to look at two different files that interact with each other.

Centril:
> Can you elaborate? Why is it the right way, or better than the current code?

The way this is encoded right now assumes that rustc's parser will always be able to parse `default fn` even on stable. That's not an assumption you should make because it will force rustc do parse `default fn` forever even if we decide to remove specialization or change the syntax somehow. By moving the use of the unstable syntax into its own `mod foobar;` and using `#[cfg(CONDITIONALLY_FALSE)]` you will make libsyntax avoid parsing the contents of `foobar` if `CONDITIONALLY_FALSE` does not hold.

emilio:
That seems like a pretty common footgun, isn't it? Is that something that rustc should warn about somehow?

Centril:
Not that common I think but I'm working on doing more pre-expansion gating so you don't get into this situation in most cases.

Issue #158, opened 2019-07-20
Why use `1usize` instead of `0usize` in unreachable?


playXE:
Why use `1usize` instead of `0usize` there :
```rust
#[inline]
pub unsafe fn unreachable() -> ! {
    enum Void {}
    let x: &Void = mem::transmute(1usize);
    match *x {}
}

```
If this function should trigger UB?

emilio:
Doesn't really matter, to be clear, right?

It creates a null reference and dereferences it, so it's clearly UB regardless of the value of the reference.

Anyhow on trunk this has been replaced by `std::hint::unreachable_unchecked`.

Issue #73, opened 2017-11-29
[meta] Wishlist for 1.0


mbrubeck:
This is a list of breaking changes that we would like to make before stabilizing the API for smallvec 1.0.

Increases the minimum Rust version from 1.20:

- [x] Use the `alloc` crate for `no_std` compatibility (#159, Requires Rust 1.36).
- [x] Replace `mem::uninitialized` with `MaybeUninit` (#126, Requires Rust 1.36).
- [x] Use `NonNull<T>` instead of `*mut T` (#171, Requires Rust 1.25).
- [x] Replace the `unreachable` crate with `std::hint::unreachable_unchecked` (#128, Requires Rust 1.27).

Breaking API changes:

- [x] Change `drain` method to take a `RangeBounds` argument (#142, Requires Rust 1.28).

Deprecations:

- [x] Remove deprecated `SmallVecN` type aliases.
- [x] Remove deprecated `push_all_move` method.
- [x] Remove deprecated `VecLike` trait.

Still depends on unstable Rust features. Postpone until 2.0:

- [ ] Use const generics to change `SmallVec<[T; N]>` to `SmallVec<T, N>`. Remove the `Array` trait. (rust-lang/rust#44580).
- [ ] Support custom allocators (#55, rust-lang/rust#32838).
- [ ] Use specialization to optimize `extend`, `insert_many`, and `from` for slices of `Copy` types. Deprecate and remove `extend_from_slice`, `insert_from_slice`, and `from_slice` (rust-lang/rust#31844).
- [ ] Use the `union` representation by default and remove the optional `union` feature added in #94 (rust-lang/rust#32836).

llogiq:
Regarding specialization, I can see how the `From` implementation can work out, but wouldn't `Extend` and `insert_many` violate coherence rules?

mbrubeck:
Specializing these would require adding an intermediate trait with a default impl and specialized impls, like libstd's [`impl Extend for Vec`](https://doc.rust-lang.org/1.28.0/src/alloc/vec.rs.html#1815-1826).

cbarrick:
Might I suggest adding an updated `drain` to the list.

`SmallVec::drain` was implemented in #26, but does not accept a range argument like `Vec::drain`. At the time, `RangeArgument` (now `RangeBounds`) had not stabilized. `RangeBounds` is now stable since Rust 1.28.

mbrubeck:
[...]

PR #162, opened 2019-09-08
Use MaybeUninit for storage of inline items

### LINKED ISSUE #126: Possible UB from use of uninitialized `[&T; N]`
Creating an array with `mem::uninitialized` may be UB if the array's elements contain non-null types (like `&T`) or uninhabited types (like `!`) or other types that have invalid values.

This can result in undefined behavior even if you never read the elements directly while they are uninitialized.  For example, this program prints "true" if optimizations are enabled, when built with current rustc:

```rust
fn main() {
    let x: Option<[&u8; 1]> = unsafe { Some(std::mem::uninitialized()) };
    println!("{}", x.is_none());
}
```

`SmallVec` isn't eligible for null pointer optimization, so it isn't affected by this.  However, it's possible that a future Rust compiler could add other optimizations that break if SmallVec uses `mem::uninitialized` to create arrays of such types.

The ideal solution is to switch from `uninitialized` to the new `MaybeUninit` union (rust-lang/rfcs#1892) when it becomes stable.

thomcc:
Fixes #126

I didn't see a PR open for this, and was poking around so I figured I'd do it. Sorry if I stepped on anyone's toes.

This should remove the use of mem::uninitialized, and ensure that all values used directly or via references are initialized. (e.g. T and &T are always initialized, but `*const T` and `*mut T` may not be, which is fine).

With these changes, `miri` passes. I also changed the CI to run `miri`, hopefully.

That said, this includes three breaking changes, so it will require a min version bump.

1. The functions on the `Array` trait `ptr` and `ptr_mut` have been removed. Because these took a `&self`/`&mut self` argument, there's no way for us to call them when we only have a `MaybeUninit<A>`. Now, we just use the memory of the object directly.

    This limits the flexibility of custom implementations of `Array`, as they can no longer return pointers to values other than themselves, but hopefully nobody does that (IMO there's a pretty good case for sealing `Array`...)

2. `SmallVec::from_buf_and_len_unchecked` now takes a MaybeUninit<A> and not an A. This means that callers have the option of only partially initializing said array without risking nasal demons. I can undo this change if desired, as it's 

3. Rust 1.36.0 is required to use MaybeUninit, so this bumps the MSRV from 1.20.0 to 1.36.0. It seems tricky to make the use of MaybeUninit optional here, at least without a bunch or work (ignoring the fact that doing so would require adding a build.rs which parses `rustc -Vv` to tell us the version).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/162)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.20.0
+  - 1.36.0
   - nightly
   - beta
   - stable
@@ -11,4 +11,5 @@ script: |
   ([ $TRAVIS_RUST_VERSION != nightly ] || cargo check --verbose --no-default-features) &&
   ([ $TRAVIS_RUST_VERSION != nightly ] || cargo test --verbose --features union) &&
   ([ $TRAVIS_RUST_VERSION != nightly ] || cargo test --verbose --all-features) &&
-  ([ $TRAVIS_RUST_VERSION != nightly ] || cargo bench --verbose bench)
+  ([ $TRAVIS_RUST_VERSION != nightly ] || cargo bench --verbose bench) &&
+  ([ $TRAVIS_RUST_VERSION != nightly ] || bash ./scripts/run_miri.sh)
--- a/lib.rs
+++ b/lib.rs
@@ -63,7 +63,7 @@ use std::iter::{repeat, FromIterator, IntoIterator};
 #[cfg(feature = "serde")]
 use std::marker::PhantomData;
 use std::mem;
-use std::mem::ManuallyDrop;
+use std::mem::MaybeUninit;
 use std::ops;
 use std::ptr;
 use std::slice;
@@ -280,29 +280,27 @@ impl<'a, T: 'a> Drop for Drain<'a, T> {
 
 #[cfg(feature = "union")]
 union SmallVecData<A: Array> {
-    inline: ManuallyDrop<A>,
+    inline: MaybeUninit<A>,
     heap: (*mut A::Item, usize),
 }
 
 #[cfg(feature = "union")]
 impl<A: Array> SmallVecData<A> {
     #[inline]
-    unsafe fn inline(&self) -> &A {
-        &self.inline
+    unsafe fn inline(&self) -> *const A::Item {
+        self.inline.as_ptr() as *const A::Item
     }
     #[inline]
-    unsafe fn inline_mut(&mut self) -> &mut A {
-        &mut self.inline
+    unsafe fn inline_mut(&mut self) -> *mut A::Item {
+        self.inline.as_mut_ptr() as *mut A::Item
     }
     #[inline]
-    fn from_inline(inline: A) -> SmallVecData<A> {
-        SmallVecData {
-            inline: ManuallyDrop::new(inline),
-        }
+    fn from_inline(inline: MaybeUninit<A>) -> SmallVecData<A> {
+        SmallVecData { inline }
     }
     #[inline]
-    unsafe fn into_inline(self) -> A {
-        ManuallyDrop::into_inner(self.inline)
+    unsafe fn into_inline(self) -> MaybeUninit<A> {
+        self.inline
     }
     #[inline]
     unsafe fn heap(&self) -> (*mut A::Item, usize) {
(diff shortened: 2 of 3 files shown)


lib.rs
```diff
@@ -403,11 +403,13 @@ impl<A: Array> SmallVec<A> {
     /// Construct an empty vector
     #[inline]
     pub fn new() -> SmallVec<A> {
-        unsafe {
-            SmallVec {
-                capacity: 0,
-                data: SmallVecData::from_inline(mem::uninitialized()),
-            }
+        debug_assert!(
```
thomcc: This is an attempt to catch any obviously invalid custom implementations of `Array`. I'd rather do this statically, but can't find a way to, since the compiler won't let me use the sizes of the generic parameters statically :(

Shnatsel:
If you're bumping MSRV, you might as well get rid of hacks such as #158

thomcc:
I saw that but didn't want to change too many unrelated things in this PR. I'm willing to make that change too if desired though.

lib.rs
danielhenrymantilla: You can go and use an `assert!` there: the condition will be trivially true or trivially false after monomorphisation, so it will get elided if the test passes

thomcc: Done.

64:
FWIW this PR fixes instances of UB in firefox, among other things.

e.g (not a comprehensive list):
* https://searchfox.org/mozilla-central/source/servo/components/style/properties/cascade.rs#245
* https://searchfox.org/mozilla-central/source/servo/components/style/properties/declaration_block.rs#1359
* https://searchfox.org/mozilla-central/source/servo/components/style/invalidation/element/invalidator.rs#102

lib.rs
```diff
@@ -403,11 +403,15 @@ impl<A: Array> SmallVec<A> {
...
-        unsafe {
-            SmallVec {
-                capacity: 0,
-                data: SmallVecData::from_inline(mem::uninitialized()),
-            }
+        // Try to detect invalid custom implementations of `Array`. Hopefuly,
+        // this check should be optimized away entirely for valid ones.
+        assert!(
+            (mem::size_of::<A>() == A::size() * mem::size_of::<A::Item>()) &&
+            (mem::align_of::<A>() >= mem::align_of::<A::Item>())
```
emilio: nit: Parens here are redundant.

jonhoo:
This would be a good change to land given the new [`mem::unitialized` lint in 1.38](https://blog.rust-lang.org/2019/09/26/Rust-1.38.0.html#linting-some-incorrect-uses-of-mem:{uninitialized,-zeroed}).

jonhoo:
@thomcc Do you have some spare cycles to look at the merge conflict?
@jdm/@emilio Is this something you'd consider merging soon? Currently, running `miri` on anything that depends on `SmallVec` gives
```
error[E0080]: Miri evaluation error: type validation failed: encountered undefined address in pointer at [0]
   --> /home/jon/.cargo/registry/src/github.com-1ecc6299db9ec823/smallvec-0.6.10/lib.rs:410:49
    |
410 |                 data: SmallVecData::from_inline(mem::uninitialized()),
    |                                                 ^^^^^^^^^^^^^^^^^^^^ Miri evaluation error: type validation failed: encountered undefined address in pointer at [0]
    |
    = note: inside call to `smallvec::SmallVec::<[&str; 1]>::new` at /home/jon/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/ops/function.rs:227:5
```

thomcc:
> @thomcc Do you have some spare cycles to look at the merge conflict?
Ah, right, thanks for the prod. Sorry about the delay, I'll update it tonight.

thomcc:
> So this makes SmallVec<[T; size]> use MaybeUninit<[T; size]>, shouldn't it conceptually use [MaybeUninit<T>; size]?

IIUC there's no difference in practice. And this is less of a change to the API surface, and less work to implement. But yeah... possibly?

> Anyhow this looks fine, with the Cargo.lock updated to account for the new minimum rust version change.

Not sure what you mean by this. This repository has Cargo.lock in it's .gitignore.

emilio:
I meant the Cargo.toml version, sorry. But I guess we can just update it later if you want.

Anyhow this looks good.

thomcc:
I've rebased the branch so that it's up to date with the latest master, but I'm... unsure what github is still upset about regarding conflicts. 😅

.travis.yml
```diff
@@ -11,4 +11,5 @@ script: |
   ([ $TRAVIS_RUST_VERSION != nightly ] || cargo check --verbose --no-default-features) &&
   ([ $TRAVIS_RUST_VERSION != nightly ] || cargo test --verbose --features union) &&
   ([ $TRAVIS_RUST_VERSION != nightly ] || cargo test --verbose --all-features) &&
-  ([ $TRAVIS_RUST_VERSION != nightly ] || cargo bench --verbose bench)
+  ([ $TRAVIS_RUST_VERSION != nightly ] || cargo bench --verbose bench) &&
+  ([ $TRAVIS_RUST_VERSION != nightly ] || bash ./scripts/run_miri.sh)
```
RalfJung: This is awesome -- I had adding Miri here on my TODO list but didn't get around to actually do that yet. :)

jonhoo:
Any chance we could do a minor release with this too so downstream projects can run miri again? :)

mbrubeck:
I would prefer to do this in the next major release, because it increases our minimum Rust version to 1.36 (from 1.20 in the previous release).  Although a Rust version increase is not a breaking change at the source level, it still causes problems for downstream crates when it happens in a minor update to a widely-used library.

#73 is tracking the breaking changes going into the next major release. I'll try to get the last items merged in the next few days, and then publish the release.

RalfJung:
When https://github.com/rust-lang/rust/pull/66059 lands, many users of the 0.6 series will see their programs break because of the incorrect `mem::uninitialized`. Is there any way a variant of this patch could land on the 0.6 branch so that getting the fix to them doesn't require a major version bump?

RalfJung:
If the concern is the MSRV bump, one option is to use https://crates.io/crates/maybe-uninit: for Rust versions prior to 1.36, that's not actually fixing the UB, but it is certainly no worse than what smallvec currently does. And for newer Rust versions, the UB is fixed.

mbrubeck:
Submitted #180 to backport the MaybeUninit changes to 0.6 using the `maybe-uninit` crate.

PR #163, opened 2019-09-20
lib.rs:  + 96-item inline support

mcronce:
Added support for `[_; 96]` arrays

I'm not sure how desirable this is to the broader user base, but I had a need for precisely this size for storing hex-encoded SHA384 hashes

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/163)


emilio:
I rebased and addressed my nit.

@bors-servo r+

PR #164, opened 2019-10-04
Deprecate unreachable function

xfix:
This is designed to merged after #162, as it deprecates a function that should be no longer necessary on newer Rust versions.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/164)

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -54,6 +54,7 @@ use std::borrow::{Borrow, BorrowMut};
 use std::cmp;
 use std::fmt;
 use std::hash::{Hash, Hasher};
+use std::hint::unreachable_unchecked;
 use std::iter::{IntoIterator, FromIterator, repeat};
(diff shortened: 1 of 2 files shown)


lib.rs
```diff
@@ -130,12 +131,11 @@ macro_rules! smallvec {
 /// Hint to the optimizer that any code path which calls this function is
 /// statically unreachable and can be removed.
 ///
-/// Equivalent to `std::hint::unreachable_unchecked` but works in older versions of Rust.
+/// Equivalent to `std::hint::unreachable_unchecked`.
 #[inline]
+#[deprecated(note = "Use std::hint::unreachable_unchecked instead")]
```
emilio: Given we're going to do a breaking version change, I'd rather just remove it.

PR #166, opened 2019-10-04
Use SliceIndex trait for SmallVec's Index trait

xfix:
Requires Rust 1.28.0.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-smallvec/166)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.20.0
+  - 1.36.0
   - nightly
   - beta
   - stable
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.20.0
```
emilio: Then can we make this check for 1.28?

xfix: My intent is to merge this alongside https://github.com/servo/rust-smallvec/pull/162, if I provided a different version, this would cause a merge conflict.

PR #171, opened 2019-10-28
Use NonNull instead of a raw pointer

mbrubeck:
This may reduce space requiments due to the null pointer optimization.


mbrubeck:
This commit was cherry-picked from #155.

PR #174, opened 2019-10-30
Update to Rust 2018 edition

mbrubeck:

PR #168, opened 2019-10-13
Add simple fuzzing

dpc:
Add simple fuzzing
    
Add infrastructure to automatically run fuzzers in CI,
and implement a simple fuzzing test based on triggering all (most)
public APIs in a randimized way.
    
As far as I was able to try, it catches the previous unsoundness issues
in a matter of seconds. This can be tried by changing the `path = "../"` dependency to
`version = "=0.6.3"` etc. and running the fuzzer manually. (Note: You'll need
to tweak the `Cargo.lock` to allow downloading the yanked versions).

PR #175, opened 2019-10-30
Version 1.0.0

mbrubeck:
* Requires Rust 1.36 or later.
* [breaking change] Use `MaybeUninit` to avoid possible undefined behavior (#162, #170).
* [breaking change] The `drain` method now takes a range argument, just like the standard `Vec::drain` (#145).
* [breaking change] Remove the `unreachable` function and replace it with the new standard `unreachable_unchecked` function (#164).
* [breaking change] Use `no_std` by default. This crate depends only on `core` and `alloc` by default. If the optional `write` feature is enabled then it depends on `std` so that `SmallVec<[u8;_]>` can implement the `std::io::Write` trait (#173).
* Add support for 96-element small vectors, `SmallVec<[T; 96]>` (#163).
* Iterators now implement `FusedIterator` (#172).
* Indexing now uses the standard `SliceIndex` trait (#166).
* Remove the deprecated `VecLike` trait (#165).
* Use `NonNull` internally (#171).
* Add automatic fuzz testing and MIRI testing (#168, #162).
* Update syntax and formatting to Rust 2018 standard (#174, #167).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "smallvec"
-version = "0.6.10"
+version = "1.0.0"
 edition = "2018"
 authors = ["Simon Sapin <simon.sapin@exyr.org>"]
 license = "MIT/Apache-2.0"


jdm:
@emilio Any feedback on this?

emilio:
I'm ok with this. Not sure if we should wait more to get to 1.0, I don't have strong opinions on that :)

jdm:
I found @mbrubeck's reasoning in https://github.com/servo/rust-smallvec/issues/73#issuecomment-548116626 compelling.

PR #180, opened 2019-11-04
Backport MaybeUninit changes to 0.6

mbrubeck:
This is a backport of #162 and #170 to the smallvec 0.6 branch. To avoid bumping the minimum Rust version, the `maybe-uninit` crate is used in place of `std::mem::MaybeUninit`.  To avoid breaking changes, the `Array::ptr` and `ptr_mut` methods are retained but are no longer used, and the API to `from_buf_and_len_unchecked` is unchanged.


RalfJung:
Awesome, thanks. :)

What does your release schedule look like? (I exempted arrays from https://github.com/rust-lang/rust/pull/66059 for now to give smallvec users some more time.)

mbrubeck:
Published version 0.6.13.

Issue #181, opened 2019-11-16
Add as_slice and as_mut_slice to smallvec::IntoIter


rklaehn:
SmallVec is mostly a drop in replacement for Vec. There are two methods on [std::vec::IntoIter](https://doc.rust-lang.org/std/vec/struct.IntoIter.html) that allow random access on the remainder of the iterator: [as_slice](https://doc.rust-lang.org/std/vec/struct.IntoIter.html#method.as_slice) and [as_mut_slice](https://doc.rust-lang.org/std/vec/struct.IntoIter.html#method.as_mut_slice). 

These are missing in SmallVec, but implementing them would be rather easy.

rklaehn:
I gave it a shot...

emilio:
Fixed by #182, thanks!

PR #182, opened 2019-11-16
Add as_slice and as_mut_slice methods

rklaehn:
To allow random access for remaining data in smallvec::IntoIter, just like
std::vec::IntoIter

Implements https://github.com/servo/rust-smallvec/issues/181


mbrubeck:
@bors-servo r+

Thanks!

PR #186, opened 2019-12-06
Miri now supports catching panics

RalfJung:
Also fix a mistake in the .gitignore (at least on my system, this didn't actually ignore the lockfile)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
 target
-./Cargo.lock
+/Cargo.lock
--- a/lib.rs
+++ b/lib.rs
@@ -15,7 +15,7 @@
 //! When this feature is enabled, `smallvec` depends on `std`.
 //!
 //! To depend on `smallvec` without `libstd`, use `default-features = false` in the `smallvec`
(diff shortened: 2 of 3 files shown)


thomcc:
This is missing https://github.com/servo/rust-smallvec/blob/master/lib.rs#L1987 too, but otherwise is awesome to see.

RalfJung:
@thomcc that test is quite odd as it is never actually enabled even without Miri! The `std` feature does not seem to exist any more, but still has some references left to it in the code. I fixed that.

However, when Miri runs `test_insert_many_panic`, it reports a memory leak: two heap allocations of size 1 get leaked. I assume that is the two `Box::new(false)` in that test. Is it expected that those get leaked?

mbrubeck:
> Is it expected that those get leaked?

Yes, we intentionally leak on panic in `insert_many`. See #96 and 26b249075930b46cfafc70b1d18fd0cb35fd2314 for details.

RalfJung:
Okay, Miri can now ignore leaks and I added that flag here. Should be good to go.

mbrubeck:
@bors-servo r+

Thanks!

PR #188, opened 2019-12-10
Link to docs.rs for documentation

mbrubeck:
because doc.servo.org might not always have the same version as crates.io


jdm:
@bors-servo r+
Thanks!

Issue #184, opened 2019-11-21
into_boxed_slice


zbraniecki:
Would it be hard to get an `into_boxed_slice` method for smallvec?

It would make replacing `Vec` for parsers which produce AST which stores `Box<[T]>` much easier.

mbrubeck:
Sounds good. If you want to submit a pull request for this, it could be implemented as `self.into_vec().into_boxed_slice()`.

L0uisc:
@mbrubeck your proposed solution will work, but won't it defeat the object of trying to replace `Vec` for parsers? I mean, then you simply end up allocating all vectors, even small ones, on the heap to be able to `Box` them. I understand that just making a slice pointing to some stack memory will be dangerous or extremely complex to keep safe, but what is the gain of using `SmallVec` then?

mbrubeck:
`Box` is always a pointer to a heap allocation, so converting a non-spilled `SmallVec` into a `Box<[T]>` *must* allocate on the heap.  If you want to avoid the heap then you should not use `Box<[T]>`.

L0uisc:
Can't you have a `Box<&T>` on the stack, pointing to a fat pointer on the heap (the actual `&T`), in turn pointing to a slice on the stack again? Anyway, I'm happy to take this up and implement your suggestion in the meantime ;)

L0uisc:
Ah, yes, I thought about `Box<[&T]>`...

PR #190, opened 2019-12-11
Implement into_boxed_slice() and doc-comment it.

L0uisc:
Implements `into_boxed_slice()` as per suggestion in #184.

I didn't write tests, since all I did was to call an existing method of `SmallVec`, `into_vec()` to create a `Vec` from it and then call a method of `alloc::vec::Vec`, `into_boxed_slice()` on the resulting `Vec`. Thus, nothing in my code needs testing, since it is trivially offloading to other methods.

I didn't, however, find a test exercising `into_vec()`, however. That is maybe something I can do? If so, can I get some pointers to where I should look for example tests and where my tests should live?

EDIT: Never mind, I forgot that unit tests are in the same file, so I scrutinized smallvec_ops.rs to search for a test for `into_vec()` and couldn't find it.

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -37,6 +37,7 @@ extern crate alloc;
 #[cfg(any(test, feature = "write"))]
 extern crate std;
 
+use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::borrow::{Borrow, BorrowMut};
 use core::cmp;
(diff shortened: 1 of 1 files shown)


mbrubeck:
@bors-servo r+

Thanks, looks good!

Issue #178, opened 2019-11-04
Implement Clone for IntoIter<A: Array> where A: Clone


robinhundt:
Implement Clone for IntoIter<A: Array> where A: Clone

This would be in line with the Iter struct that is returned when calling `into_iter()` on a `Vec`.
https://doc.rust-lang.org/std/slice/struct.Iter.html

mbrubeck:
If someone would like to implement this, it could be done using `IntoIter::as_slice` (recently added in #182) and `SmallVec::from_slice`.

PR #191, opened 2019-12-12
Implement Clone for IntoIter where A: Clone

L0uisc:
Implements `Clone` for `IntoIter<A>` where `A: Array + Clone` and `A::Item: Copy` (necessary to be able to use `from_slice()`) as per #178.

I also left a commented implementation which requires `A::Item` to be the more relaxed `Clone`, which also works. However, it will be less efficient than the other impl when the types are `Copy` because it can't use `from_slice()`. I believe the feature I'm looking for is specialization?


mbrubeck:
Ah, I forgot that `from_slice` requires `Copy`.

I think we should use the more-general, less-performant version for now, and when specialization is available we can use it to optimize the `Copy` case.

If you use `SmallVec::from(slice)` instead of `SmallVec::from_slice(slice)` then this should work for `Clone` types, and when the `"specialization"` Cargo feature is enabled it will use `from_slice` when possible.

L0uisc:
Sorry, I'm trying out Visual Studio's git plugin, and it seems I'm doing things wrong...

L0uisc:
#192 should be correct.

PR #192, opened 2019-12-12
Implement Clone for IntoIter<A> where A: Clone

### LINKED ISSUE #178: Implement Clone for IntoIter<A: Array> where A: Clone

L0uisc:
Implemented code to fix #178. The commented impl can be uncommented once specialization is possible. Both impls were tested and passed the three added tests.

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -1504,15 +1504,6 @@ pub struct IntoIter<A: Array> {
     end: usize,
 }
 
-/*impl<A: Array + Clone> Clone for IntoIter<A>
-where
-    A::Item: Copy,
-{
-    fn clone(&self) -> IntoIter<A> {
-        SmallVec::from_slice(self.as_slice()).into_iter()
-    }
-}*/
-
 impl<A: Array + Clone> Clone for IntoIter<A>
 where
     A::Item: Clone,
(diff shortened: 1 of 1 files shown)


lib.rs
```diff
@@ -1504,6 +1504,24 @@ pub struct IntoIter<A: Array> {
     end: usize,
 }
 
+/*impl<A: Array + Clone> Clone for IntoIter<A>
```
mbrubeck: You can remove this. Since we already have a specialization for `SmallVec::from`, we won't need a separate specialized impl here.

L0uisc: I'm just having a doubt about this: if I just keep the below impl, won't `IntoIter<A>::clone()` only work where `A::Item: Clone`? I don't know how Rust plans to implement specialization, but won't the impl which limits `A::Item` to types which implement `Clone` then prevent `Copy` types to call the less-expensive `From` impl?

L0uisc: Or do you mean we just provide one impl and modify it when const generics land to work well with all possible `From` implementations?

mbrubeck: Right.

PR #195, opened 2020-01-08
feat: Add splice

Marwes:
To mirror https://doc.rust-lang.org/alloc/vec/struct.Vec.html#method.splice .

The implementation is a straight copy of `Vec`s implementation, only
altered to call `len()/reserve()` instead of accessing the internal
`len` and `buf` field respectively as those do not exist in `SmallVec`.
An extra type parameter were also required for `Drain`.

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -251,6 +251,117 @@ impl<'a, T: 'a + Array> Drop for Drain<'a, T> {
     }
 }
 
+/// A splicing iterator for `SmallVec`.
+///
+/// This struct is created by the [`splice()`] method on [`SmallVec`]. See its
+/// documentation for more.
+pub struct Splice<'a, I: Iterator<Item = A::Item> + 'a, A: Array + 'a> {
+    drain: Drain<'a, A>,
+    replace_with: I,
+}
+
+impl<I: Iterator<Item = A::Item>, A: Array> Iterator for Splice<'_, I, A> {
+    type Item = I::Item;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.drain.next()
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.drain.size_hint()
+    }
+}
+
+impl<I: Iterator<Item = A::Item>, A: Array> DoubleEndedIterator for Splice<'_, I, A> {
+    fn next_back(&mut self) -> Option<Self::Item> {
+        self.drain.next_back()
+    }
+}
+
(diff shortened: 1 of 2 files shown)


Marwes:
Copied over the tests for `Vec::splice`, those don't test the additional edge conditions that `SmallVec` has but its a start.

emilio:
Seems there's something wrong as tests are failing. The error looks legit:

```
error: Miri evaluation error: Memory access failed: pointer must be in-bounds at offset 24, but is outside bounds of allocation 25122 which has size 20
    --> /home/travis/.rustup/toolchains/nightly-2020-01-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/intrinsics.rs:1586:5
     |
1586 |     copy(src, dst, count)
     |     ^^^^^^^^^^^^^^^^^^^^^ Miri evaluation error: Memory access failed: pointer must be in-bounds at offset 24, but is outside bounds of allocation 25122 which has size 20
     |
     = note: inside call to `std::intrinsics::copy::<i32>` at /home/travis/build/servo/rust-smallvec/lib.rs:360:9
     = note: inside call to `smallvec::Drain::<[i32; 1]>::move_tail` at /home/travis/build/servo/rust-smallvec/lib.rs:302:17
     = note: inside call to `<smallvec::Splice<std::iter::Cloned<std::slice::Iter<i32>>, [i32; 1]> as std::ops::Drop>::drop` at /home/travis/.rustup/toolchains/nightly-2020-01-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/ptr/mod.rs:184:1
note: inside call to `std::ptr::real_drop_in_place::<smallvec::Splice<std::iter::Cloned<std::slice::Iter<i32>>, [i32; 1]>> - shim(Some(smallvec::Splice<std::iter::Cloned<std::slice::Iter<i32>>, [i32; 1]>))` at tests/splice.rs:7:38
    --> tests/splice.rs:7:38
     |
7    |     v.splice(2..4, a.iter().cloned());
     |                                      ^
note: inside call to `test_splice` at tests/splice.rs:4:1
    --> tests/splice.rs:4:1
     |
4    | / fn test_splice() {
5    | |     let mut v: SmallVec<[_; 1]> = smallvec![1, 2, 3, 4, 5];
6    | |     let a = [10, 11, 12];
7    | |     v.splice(2..4, a.iter().cloned());
...    |
10   | |     assert_eq!(&v[..], &[1, 20, 11, 12, 5]);
11   | | }
     | |_^
```

PR #196, opened 2020-01-24
impl Debug for IntoIter

Nemo157:
This has long been provided by `std::vec::IntoIter`, so is useful in cases where it is being directly replaced.


mbrubeck:
@bors-servo r+

Thanks!

PR #199, opened 2020-01-28
Version 1.2.0

mbrubeck:
Changes in this release:

* `IntoIter` now implements `Debug` (#196).
* `smallvec!` macro is now easier to use in `no_std` contexts where the `vec!` macro isn't automatically imported (#198).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "smallvec"
-version = "1.1.0"
+version = "1.2.0"
 edition = "2018"
 authors = ["Simon Sapin <simon.sapin@exyr.org>"]
 license = "MIT/Apache-2.0"


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "smallvec"
-version = "1.1.0"
+version = "1.2.0"
```
emilio: Hmm, actually, does this need to be a breaking release?

mbrubeck: In [semver](https://semver.org/), for stable releases (version >= 1.0.0), the first number (MAJOR) is incremented for breaking changes. The second number (MINOR) is incremented for additional features added without breaking backward compatibility. The third number (PATCH) is incremented for bug fixes that do not add new API or break existing API.

Cargo treats two versions as "compatible" if the first *non-zero* component is the same. So 0.1.0 to 0.2.0 is a breaking change, but 1.1.0 to 1.2.0 is not.

emilio:
Ah, nice, I always forget.

@bors-servo r+

PR #203, opened 2020-02-21
added to_smallvec extension trait to slice primitive

### LINKED ISSUE #116: Add to_smallvec for the slice primitive type
The slice primitive type has a [`to_vec` method](https://doc.rust-lang.org/std/primitive.slice.html#method.to_vec) which makes it convenient to convert a slice to a vector. It would be great if this crate could define an extension trait that added a method called `to_smallvec` (or some other name) which did the same thing for `SmallVec`.

This makes it particularly convenient to port code to using SmallVec since you can just import that extension trait instead of converting all the calls to `to_vec`. :smiley:

UnicodingUnicorn:
Should resolve #116.

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -1684,6 +1684,17 @@ impl_array!(
     0x40000, 0x60000, 0x80000, 0x10_0000
 );
 
+trait ToSmallVec<A:Array> {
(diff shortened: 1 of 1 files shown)


mbrubeck:
@bors-servo r+

Thanks!

PR #204, opened 2020-03-17
Add support for constant generics

c410-f3r:
No breaking changes and no internal hacks. This PR only adds a feature to switch between implementations of the `Array` trait.

I personally think that this approach is pretty reasonable and should be at least considered for discussion.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ readme = "README.md"
 documentation = "https://docs.rs/smallvec/"
 
 [features]
+const_generics = []
 write = []
 union = []
 specialization = []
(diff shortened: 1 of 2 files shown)


mbrubeck:
Cool, thanks! It looks like the `.travis.yml` file needs some updates to make the tests pass in CI, but otherwise this looks great.

c410-f3r:
Thank you! Thank you very much!

mbrubeck:
Let me know if you need this published to crates.io soon; otherwise we will probably wait until we have a few more changes before making a new release.

c410-f3r:
I am fine using the git version and now that this change is upstream, everything is cool. Thank you one more time.

PR #206, opened 2020-04-04
Add #[inline] attribute to all fns which return SmallVec

hiddenhare:
When rustc fails to inline a `SmallVec` constructor, it can carry a significant performance cost: for example, if `SmallVec::<[i64; 128]>::from_iter(...)` fails to inline, it will perform an unnecessary 1kb memcpy.

I've recently been bitten by this when using `SmallVec` in a context where rustc seemed to be reluctant to perform inlining (a large fn with nested closures). Switching from `SmallVec::from_iter` to `SmallVec::new` and `push`, with a buffer size of 256 bytes, saved over 20ns per call. For larger buffers, `from_iter` carried a proportionally higher cost, even when the actual capacity in use didn't change.

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -1055,6 +1055,7 @@ impl<A: Array> SmallVec<A> {
     ///         assert_eq!(&*rebuilt, &[4, 5, 6]);
     ///     }
     /// }
+    #[inline]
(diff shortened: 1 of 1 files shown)


mbrubeck:
@bors-servo r+

Thank you!

Issue #208, opened 2020-04-10
Panicking insert_many leaks memory


RalfJung:
The `test_insert_many_panic` test leaks memory. Both of these allocations here to do get freed, presumably due to the panicking `insert_many`:

https://github.com/servo/rust-smallvec/blob/3957cd8b36665e8199ac76f7e50c37a23d658a51/lib.rs#L2096-L2101

Found by Miri:
```
note: tracking was triggered
    --> /home/r/.rustup/toolchains/miri/lib/rustlib/src/rust/src/liballoc/alloc.rs:80:47
     |
80   |     __rust_alloc(layout.size(), layout.align())
     |                                               ^ created allocation with id 129830
     |
     = note: inside `alloc::alloc::alloc` at /home/r/.rustup/toolchains/miri/lib/rustlib/src/rust/src/liballoc/alloc.rs:80:47
     = note: inside `<alloc::alloc::Global as core::alloc::AllocRef>::alloc` at /home/r/.rustup/toolchains/miri/lib/rustlib/src/rust/src/liballoc/alloc.rs:174:49
     = note: inside `alloc::alloc::exchange_malloc` at /home/r/.rustup/toolchains/miri/lib/rustlib/src/rust/src/liballoc/alloc.rs:268:11
     = note: inside `alloc::boxed::Box::<bool>::new` at /home/r/.rustup/toolchains/miri/lib/rustlib/src/rust/src/liballoc/boxed.rs:174:9
note: inside `tests::test_insert_many_panic` at lib.rs:2100:26
    --> lib.rs:2100:26
     |
2100 |                 dropped: Box::new(false),
     |                          ^^^^^^^^^^^^^^^
note: inside closure at lib.rs:2072:5
    --> lib.rs:2072:5
     |
2072 | /     fn test_insert_many_panic() {
2073 | |         struct PanicOnDoubleDrop {
2074 | |             dropped: Box<bool>,
2075 | |         }
...    |
2107 | |         assert!(result.is_err());
2108 | |     }
     | |_____^

The following memory was leaked:
alloc129816 (Rust heap, size: 1, align: 1) {
    00                                              │ .
}
alloc129830 (Rust heap, size: 1, align: 1) {
    00                                              │ .
}
```

mbrubeck:
This leak was intentional; see #96 and https://github.com/servo/rust-smallvec/commit/26b249075930b46cfafc70b1d18fd0cb35fd2314 for details.

[...]

PR #209, opened 2020-04-11
enable Miri leak checker

RalfJung:
Works around https://github.com/servo/rust-smallvec/issues/208

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -865,6 +865,8 @@ impl<A: Array> SmallVec<A> {
 
     /// Insert multiple elements at position `index`, shifting all following elements toward the
     /// back.
+    ///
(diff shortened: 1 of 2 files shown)


lib.rs
```diff
@@ -865,6 +865,8 @@ impl<A: Array> SmallVec<A> {
 
     /// Insert multiple elements at position `index`, shifting all following elements toward the
     /// back.
+    ///
+    /// Note: when the iterator panics, this can leak memory.
```
RalfJung, line 869: I added this when I read that the leak was intentional; if this is still considered a bug I can also remove it again.

mbrubeck: Let's keep it for now, and remove it if/when the leak is fixed.

PR #213, opened 2020-04-22
Fix leak on panic in `insert_many`.

### LINKED ISSUE #208: Panicking insert_many leaks memory

mbrubeck:
Fixes #208 and reverts the workaround added in #209. CC @RalfJung.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -98,7 +98,7 @@ use core::hint::unreachable_unchecked;
 use core::iter::{repeat, FromIterator, FusedIterator, IntoIterator};
 use core::mem;
 use core::mem::MaybeUninit;
-use core::ops::{self, RangeBounds};
+use core::ops::{self, Range, RangeBounds};
 use core::ptr::{self, NonNull};
 use core::slice::{self, SliceIndex};
 
@@ -975,8 +975,6 @@ impl<A: Array> SmallVec<A> {
 
     /// Insert multiple elements at position `index`, shifting all following elements toward the
     /// back.
-    ///
-    /// Note: when the iterator panics, this can leak memory.
     pub fn insert_many<I: IntoIterator<Item = A::Item>>(&mut self, index: usize, iterable: I) {
         let iter = iterable.into_iter();
         if index == self.len() {
@@ -991,13 +989,19 @@ impl<A: Array> SmallVec<A> {
         unsafe {
             let old_len = self.len();
             assert!(index <= old_len);
-            let mut ptr = self.as_mut_ptr().add(index);
+            let start = self.as_mut_ptr();
+            let mut ptr = start.add(index);
 
             // Move the trailing elements.
             ptr::copy(ptr, ptr.add(lower_size_bound), old_len - index);
 
             // In case the iterator panics, don't double-drop the items we just copied above.
-            self.set_len(index);
+            self.set_len(0);
+            let mut guard = DropOnPanic {
+                start,
+                skip: index..(index + lower_size_bound),
+                len: old_len + lower_size_bound,
+            };
 
             let mut num_added = 0;
             for element in iter {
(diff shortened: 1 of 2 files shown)


lib.rs
```diff
@@ -898,10 +901,17 @@ impl<A: Array> SmallVec<A> {
                     ptr = self.as_mut_ptr().add(index);
                     cur = ptr.add(num_added);
                     ptr::copy(cur, cur.add(1), old_len - index);
+
+                    guard.ptr = self.as_mut_ptr();
```
RalfJung, line 905: I am actually somewhat surprised that this does not invalidate `ptr`... doesn't `self.as_mut_ptr()` create a mutable reference that invalidates all other pointers? I suppose this function is called via the `DerefMut` impl, so it even creates an intermediate `&mut [Item]` that covers all elements currently marked as initialized.

mbrubeck: `len` has been set to zero at this point, so `self` currently derefs to an empty slice.

However, MIRI also does not seem to flag use use of `vec.deref().as_mut_ptr()` to create and use aliasing pointers to elements of non-empty Vecs, so maybe this is just a limitation of MIRI.

RalfJung: `Vec` has a pointer indirection, so `&mut Vec` just demands exclusive access to the fields of that struct. But I thought smallvec stores (some) elements inline, thus `&mut SmallVec` *does* demand exclusive access to those elements.

> self.len() has been set to zero at this point, so self currently derefs to an empty slice.

Well, but `&mut self` is still being passed fully to `deref_mut`, which covers all inline elements.

Also, an empty slice turned into a raw pointer should only be used to access elements of that slice (the pointer "comes from" the slice and this is restricted to the memory the slice covers). Probably there are just so many raw pointers everywhere here that Miri is fine, but that might change once we experiment with tracking raw pointers more precisely.

Even if nothing goes wrong here (I'd need to dig deeper to find out why), maybe it would be better to follow `Vec` and [have raw pointer methods that do not go through slices](https://github.com/rust-lang/rust/pull/61114)?

mbrubeck: Thanks, that makes sense. I'll try converting this to replace unique borrows with raw-pointer-only methods.

mbrubeck: Updated the code to avoid invalidating raw pointers before using them, and shadowed the `as_ptr` and `as_mut_ptr` methods (as in `std::vec::Vec`) to avoid dereferencing to slice while the contents may be invalid.

RalfJung: I don't have time to review this in detail right now but avoiding the intermediate slice is certainly an improvement. At some point I'll make Miri's raw ptr checking more strict and then we'll see what happens here and all over the rest of the ecosystem. ;)

mbrubeck:
Rebased and added some more test cases.  r? @SimonSapin or @jdm

jdm:
@bors-servo r+
Thanks for fixing this!

PR #214, opened 2020-04-22
Add `try_reserve` and friends

SimonSapin:
This more cleanly replaces this hack adding fallible allocation from outside the crate:

https://github.com/servo/servo/blob/faa9dccfe8/components/fallible/lib.rs#L101-L163

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -40,7 +40,7 @@
 //!
 //! When this feature is enabled, `SmallVec` works with any arrays of any size, not just a fixed
 //! list of sizes.
-//! 
+//!
 //! ### `specialization`
 //!
 //! **This feature is unstable and requires a nightly build of the Rust toolchain.**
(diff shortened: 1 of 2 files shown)


lib.rs
```diff
@@ -764,21 +792,30 @@ impl<A: Array> SmallVec<A> {
...
     ///
     /// Panics if the new capacity overflows `usize`.
     pub fn reserve_exact(&mut self, additional: usize) {
+        infallible(self.try_reserve_exact(additional))
+    }
+
+    /// Reserve the minimum capacity for `additional` more elements to be inserted.
+    pub fn try_reserve_exact(&mut self, additional: usize) -> Result<(), CollectionAllocErr> {
         let (_, &mut len, cap) = self.triple_mut();
         if cap - len < additional {
```
emilio, line 820: May be a bit nicer to invert the condition and early-return, then deindent the rest, but nbd.

lib.rs
```diff
@@ -764,21 +792,30 @@ impl<A: Array> SmallVec<A> {
             let new_cap = len
                 .checked_add(additional)
                 .and_then(usize::checked_next_power_of_two)
-                .unwrap_or(usize::max_value());
-            self.grow(new_cap);
+                .ok_or(CollectionAllocErr::CapacityOverflow)?;
+            self.try_grow(new_cap)
```
emilio, line 804: Here too.

PR #215, opened 2020-04-24
Split out some cold paths

SimonSapin:


SimonSapin:
@emilio Do you think this would help with https://bugzilla.mozilla.org/show_bug.cgi?id=1632469#c3 ?

lib.rs
```diff
@@ -800,8 +810,19 @@ impl<A: Array> SmallVec<A> {
...
         if cap - len >= additional {
-            return Ok(());
+            Ok(())
+        } else {
+            self.try_reserve_cold(len, additional)
         }
+    }
+
+    #[cold]
+    #[inline(never)]
```
emilio, line 820: not sure if it's worth the `#[inline(never)]`, tbf.

lib.rs
```diff
@@ -800,8 +810,19 @@ impl<A: Array> SmallVec<A> {
...
+    fn try_reserve_cold(
```
emilio, line 821: I'm not really sure about this being tagged as `#[cold]` unconditionally. There's I suspect a somewhat common usage of:

```
let mut vec = SmallVec::new();
vec.reserve(num_elements);

for .... {
  vec.push(..);
}
```

For such cases `#[cold]` is not accurate, and `#[inline(never)]` is probably counter-productive.

I think only the caller knows whether the capacity is going to be likely-already enough. For example `push()` could use the `try_reserve_cold` without much trouble.

Maybe we should expose both, or something...

lib.rs
```diff
@@ -221,12 +221,22 @@ impl From<LayoutErr> for CollectionAllocErr {
         CollectionAllocErr::CapacityOverflow
     }
 }
+impl CollectionAllocErr {
+    #[cold]
+    #[inline(never)]
+    fn bail(&self) -> ! {
```
emilio, line 227: `#[cold]` here looks uncontroversial to me, though maybe the `#[inline(never)]` prevents stuff from getting potentially optimized away?

PR #216, opened 2020-04-25
Implement Clone using From

Alexendoo:
This allows using the specialization in clones when it's enabled


mbrubeck:
@bors-servo r+

Thanks!

PR #218, opened 2020-04-30
Add resize_with

l0calh05t:
`std::vec::Vec` added `resize_with` in Rust 1.33. This patch adds the equivalent to `SmallVec`.


mbrubeck:
@bors-servo r+

Thank you!

PR #221, opened 2020-05-06
Make sure smallvec! is usable without being in scope

cuviper:
Macro recursion should use a `$crate` prefix to reliably find itself.


mbrubeck:
@bors-servo r+

Thanks!

PR #228, opened 2020-07-06
Don’t heap-allocate for zero-size items

SimonSapin:
Allocating zero bytes is Undefined Behavior.
CC https://github.com/servo/servo/pull/26304#issuecomment-653626058

### DIFF
--- a/lib.rs
+++ b/lib.rs
@@ -462,8 +462,8 @@ unsafe impl<A: Array + Sync> Sync for SmallVecData<A> {}
 /// ```
 pub struct SmallVec<A: Array> {
     // The capacity field is used to determine which of the storage variants is active:
-    // If capacity <= A::size() then the inline variant is used and capacity holds the current length of the vector (number of elements actually in use).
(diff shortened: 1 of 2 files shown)


lib.rs
```diff
@@ -611,10 +611,30 @@ impl<A: Array> SmallVec<A> {
...
+            //
+            // In our case, this also ensures that a smallvec of zero-size items never spills,
+            // and we never try to allocate zero bytes which `std::alloc::alloc` disallows.
+            core::usize::MAX
+        }
+    }
+
     /// The maximum number of elements this vector can hold inline
     #[inline]
     pub fn inline_size(&self) -> usize {
```
SimonSapin, line 636: This should not take `&self`, but it’s public so meh

SimonSapin:
Published https://crates.io/crates/smallvec/1.4.1 and yanked https://crates.io/crates/smallvec/1.4.0 since it can trigger UB from safe code.

PR #234, opened 2020-09-08
Implement Array for all sizes <= 32

thomcc:
Most traits supported by arrays are fully supported for arrays in the Rust stdlib as for all sizes <= 32. I was surprised to find that `SmallVec<[T; 18]>` (I happen to know that 18 the maximum outside of obscure cases) isn't possible because smallvec::Array is not consistent with this pattern.

Of course, I don't actually need this — I can just choose another size, but it seems like it would be good/unsurprising to be a superset of libstd's choices here.


mbrubeck:
@bors-servo r+

Thanks!

PR #238, opened 2020-11-08
Return allocation error in deserialize instead of panicking

epilys:
There's no way to catch allocation errors since out of memory errors
cause an abort. Fail gracefully by returning the error instead of
panicking.

I happened upon this error when deserializing untrusted data with bincode. Bincode provides a byte limit bound but for sequences it's not possible to enforce this through serde since collection types like smallvec handle their own allocation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -233,6 +233,12 @@ pub enum CollectionAllocErr {
     },
 }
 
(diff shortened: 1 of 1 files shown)


mbrubeck:
@bors-servo r+

Thanks!

PR #241, opened 2020-11-30
Remove extraneous branch from push

mpdn:
`push` does two branches on the "smallness" of the `smallvec`: one before the reserve check and one after. LLVM doesn't seem to optimize the second branch away for the (very common) non-growing case. In addition, in the growing branch we know the memory will be on the heap, so no need to branch here.

On my machine, this improves `bench_push` from approx 300ns to 263ns (+/- 5 on both).


mbrubeck:
@bors-servo r+

Thanks!

Issue #240, opened 2020-11-18
RFC: rewrite based on min_const_generics for 2.0


SimonSapin:
There is now a serious proposal and plan to stabilize `#![feature(min_const_generics)]` in Rust 1.50: https://github.com/rust-lang/rust/pull/79135. Using it here and removing the `Array` trait would a welcome simplification but obviously a breaking API change, so it would need to be part of a version 2.0 of the crate.

I’m wondering whether it would make sense to try a complete rewrite of `smallvec` instead of adapting the existing code. Some ideas:

* The current `capacity` field has a misleading name and non-obvious meaning, it can represent either the capacity *or* the length of the vector. I feel this makes the code harder to understand and keep correct than necessary. Instead we could store a length in the `0..isize::MAX` range, and pack a tag in the remaining bit.

* The dual representation with conditional compilation can be skipped/removed entirely, as `union`s with `ManuallyDrop<T>` fields with `T: !Copy` are now stable: https://github.com/rust-lang/rust/pull/77547

* The "core" of the crate can be rather small (and in a separate module and file), with everything else mostly copied from `vec.rs` and `raw_vec.rs` from the standard library

* Copying from the standard library again gets us more recent improvements, such as making some code less generic to reduce monomorphization bloat.

* This would be a good time to look at `unsafe` code

* 2.0 can be an opportunity to remove some APIs. For example, `IntoIterator for SmallVec` implies an expensive move. We could omit it and document to use `drain` instead.

I’d consider *not* necessarily including everything in https://github.com/servo/rust-smallvec/issues/183, as specialization and custom allocators are likely not gonna be stable before or soon after `min_const_generics` is. Adding those *might* not require breaking API changes, though?

I’ve started playing with this a little. Here is the core data structure:

```rust
pub struct SmallVec<Item, const INLINE_CAPACITY: usize> {
    // Safety invariants:
    // * The active union field of `storage` must be consistent with the tag of `tagged_len`.
    // * The length in `tagged_len` must not be greater than storage capacity
    // * The first `length` items of the storage must be initialized.
    tagged_len: TaggedLen,
    storage: Storage<MaybeUninit<Item>, INLINE_CAPACITY>,
}

/// Upper bit: 0 for inline storage, 1 for heap storage
/// The rest of the bits: SmallVec length = number of items initialized, at the start of storage.
pub(crate) struct TaggedLen(usize);

/// An array, either inline (of statically-fixed size) or heap-allocated (of dynamic size).
[...]

PR #242, opened 2020-12-02
Remove `min_const_generics` feature

c410-f3r:
Depends on https://github.com/rust-lang/rust/pull/79135.

If #240 is already under development and will be available before the 1.50 release, then feel free to close this PR. Otherwise, the feature removal will benefit upstream projects in the meanwhile.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -75,7 +75,6 @@
 #![cfg_attr(feature = "specialization", allow(incomplete_features))]
 #![cfg_attr(feature = "specialization", feature(specialization))]
 #![cfg_attr(feature = "may_dangle", feature(dropck_eyepatch))]
-#![cfg_attr(feature = "const_generics", feature(min_const_generics))]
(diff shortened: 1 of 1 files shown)


mbrubeck:
Thanks!

We will need to wait until rust-lang/rust#79135 is merged and released to the nightly channel before we can merge this PR.

src/lib.rs
```diff
@@ -1934,34 +1933,13 @@ impl<'a> Drop for SetLenOnDrop<'a> {
     }
 }
 
-#[cfg(feature = "const_generics")]
```
mbrubeck, line 1937: I'd like to keep use of const generics behind an optional feature in smallvec 1.x, so it can continue compiling on Rust versions older than 1.51.  Let's wait until 2.0 to remove this feature completely.

c410-f3r: Ooopsss... My bad :p

Issue #247, opened 2020-12-30
Compile union feature on 1.50 via ManuallyDrop?


saethlin:
On the 1.50 beta if I delete the `cfg_attr` for `feature = "union"`, I get this error message:
```rust
error[E0658]: unions with non-`Copy` fields other than `ManuallyDrop<T>` are unstable
   --> src/lib.rs:354:5
    |
354 |     inline: MaybeUninit<A>,
    |     ^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #55149 <https://github.com/rust-lang/rust/issues/55149> for more information
```
It doesn't take a lot to make the 1.50 (currently beta) compiler accept the code, but I'm a bit out of my depth here. Can we have the union feature on stable starting in 1.50?

mbrubeck:
Yes, it looks like this works!  Please feel free to submit a pull request if you have a version that compiles already.  (Or I can do it, if you prefer.)  You can also change the documentation comment at the top of `src/lib.rs` to say that the `union` feature requires Rust 1.50.

saethlin:
Closing, PR addressing this was merged

PR #248, opened 2020-12-31
Wrap with ManuallyDrop so that the union feature works on 1.50

saethlin:
Per https://github.com/servo/rust-smallvec/issues/247#issuecomment-752794872

The unrelated diff was produced by rustfmt, which my editor runs automatically. I'm a bit surprised, usually when this produces a diff it's a big one. Do you want me to revert those lines?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,7 +28,7 @@
 //!
 //! ### `union`
 //!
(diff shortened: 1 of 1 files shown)


mbrubeck:
@bors-servo r+

Thanks!  The rustfmt changes are fine, too.

PR #249, opened 2020-12-31
Version 1.6.0

mbrubeck:
Release notes:

* The `union` feature is now compatible with stable Rust 1.49 (#248, #247).
* Fixed warnings when compiling with Rust 1.51 nightly (#242, #246).


mbrubeck:
This also corrects the documentation added in #248, since the union feature is stable starting in Rust 1.49 (which releases tomorrow), not 1.50.

PR #254, opened 2021-01-08
Fix potential buffer overflow in `insert_many`

### LINKED ISSUE #252: Buffer overflow in `insert_many()`

mbrubeck:
Fixes #252.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1009,7 +1009,7 @@ impl<A: Array> SmallVec<A> {
     /// Insert multiple elements at position `index`, shifting all following elements toward the
     /// back.
     pub fn insert_many<I: IntoIterator<Item = A::Item>>(&mut self, index: usize, iterable: I) {
-        let iter = iterable.into_iter();
+        let mut iter = iterable.into_iter();
         if index == self.len() {
             return self.extend(iter);
         }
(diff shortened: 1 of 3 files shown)


mbrubeck:
Pushed a different version of the fix.  This is a more invasive change, but it leaves the code a bit simpler and much easier to reason about.

The key observation is that the guard is no longer needed after we fill up the initial `lower_size_bound` slots of reserved space.  While looping through any surplus items from the iterator, the SmallVec is in a valid state at the start and end of each iteration.

src/lib.rs
```diff
@@ -1066,12 +1058,26 @@ impl<A: Array> SmallVec<A> {
                 );
             }
 
+            // There are no more duplicate or uninitialized slots, so the guard is not needed.
             self.set_len(old_len + num_added);
+            mem::forget(guard);
+        }
+
+        unsafe {
+            for element in iter {
```
emilio, line 1067: Can this be just:

```
for element in iter {
    self.insert(index + num_added, element);
    num_added += 1;
}
```

And remove the unsafe block?

emilio (approved):
r=me with that change unless I'm missing something.

src/lib.rs
mbrubeck: Great idea.  Done.

PR #259, opened 2021-03-21
Panic on arithmetic overflow in drain

### LINKED ISSUE #258: Potential integer overflow in drain leading to draining thinks when it should be out of bounds.
The `drain` implementation can have a integer overflow in both of the match expressions.

https://github.com/servo/rust-smallvec/blob/4e53e072808815ed6b847c77193c568ee076c29d/src/lib.rs#L725-L738

This lead to possible cases where code which should panic due to being called with out of bounds indices
instead draining "something".

While it is implicitly caught on test build by the integer overflow checks it is in my opinion still a bug as no one can know that drain relies on integer overflow checks internally to assure correctness.

## Example of Bug (needs to be run in release mode)

```rust
use std::ops::{Bound, RangeBounds};

struct RangeFromExcluded {
    from: usize,
}

impl RangeBounds<usize> for RangeFromExcluded {
    fn start_bound(&self) -> std::ops::Bound<&usize> {
        Bound::Excluded(&self.from)
    }

    fn end_bound(&self) -> std::ops::Bound<&usize> {
        Bound::Unbounded
    }
}

fn main() {
    let mut v: SmallVec<[u8; 8]> = smallvec![1,2,3];

    //should panic but drains 0..0
    let _ = v.drain(..=usize::MAX);
    dbg!(&v);

    //should panic but drains full range
    let _ = v.drain(RangeFromExcluded { from: usize::MAX });
    dbg!(&v);
}
```

Once `https://github.com/rust-lang/rust/issues/76393` it could be used instead.
Until then using a `checked_add` would make sense I think.

mbrubeck:
Fixes #258.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -725,11 +725,11 @@ impl<A: Array> SmallVec<A> {
         let len = self.len();
         let start = match range.start_bound() {
             Included(&n) => n,
-            Excluded(&n) => n + 1,
+            Excluded(&n) => n.checked_add(1).expect("Range start out of bounds"),
             Unbounded => 0,
         };
         let end = match range.end_bound() {
-            Included(&n) => n + 1,
+            Included(&n) => n.checked_add(1).expect("Range end out of bounds"),
             Excluded(&n) => n,
             Unbounded => len,
         };
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -424,6 +424,13 @@ fn test_invalid_grow() {
 }
 
 #[test]
+#[should_panic]
+fn drain_overflow() {
+    let mut v: SmallVec<[u8; 8]> = smallvec![0];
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -725,11 +725,11 @@ impl<A: Array> SmallVec<A> {
         let len = self.len();
         let start = match range.start_bound() {
             Included(&n) => n,
-            Excluded(&n) => n + 1,
+            Excluded(&n) => n.checked_add(1).unwrap(),
```
emilio, line 728: May be worth adding a panic message?

emilio (approved):
Either way looks good.

src/lib.rs
mbrubeck: Added panic messages.

Issue #263, opened 2021-06-30
Make `SmallVec::new` const


Rua:
I don't know if this is possible, but it would definitely be helpful!

mbrubeck:
This should be easier to implement after #240 is complete.  I have added this to our roadmap for smallvec 2.0 (#183).

PR #265, opened 2021-09-24
Added feature `const_new` which enables `SmallVec::new_const()`

TimLuq:
As I wasn't alone in finding a use for having a `const` version of `new` (see #263) I decided to make a PR for it.

This adds a new feature flag `const_new` (nightly only) which exposes an alternative `new` function: `pub const unsafe fn new_const() -> SmallVec<A>`.

Nightly is needed for the binding `A: Array`, which could be worked around if the strict requirement is dropped from the `struct` declarations and moved only to the existing `impl`s.

If the maintainers wish for another solution for this I'll be happy to make a few more commits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ documentation = "https://docs.rs/smallvec/"
 
 [features]
 const_generics = []
+const_new = []
 write = []
 union = []
 specialization = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ documentation = "https://docs.rs/smallvec/"
 
 [features]
 const_generics = []
-const_new = []
+const_new = ["const_generics"]
 write = []
 union = []
 specialization = []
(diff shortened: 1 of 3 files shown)


TimLuq:
I added `from_const` which behaves like `from_buf` for arrays (`[_; N]`). I also added a macro `smallvec_inline` that's just sugar for `from_const`.

My thought was that `new_const` and `from_const` should be visible in the docs. So the most recent commit also includes the changes necessary for displaying feature hints on docs.rs.

mbrubeck:
Could `new_const` be made safe by using const generics to implement it only for `SmallVec<[T; N]>`?

TimLuq:
> Could `new_const` be made safe by using const generics to implement it only for `SmallVec<[T; N]>`?

Yes, that is another option that should work. It will then be added to the same place as `from_const` which seems reasonable.

I'll be able to push a new commit in 8 hours or so. Is there anything else I should take a look at regarding this PR?

TimLuq:
Limiting the const functions to `SmallVec<[T; N]>` also got rid of the need for the trait bound. This means it now builds in stable 1.51+ with the `const_new` feature active.

mbrubeck:
Thanks!  This has been released in smallvec 1.7.0: https://crates.io/crates/smallvec/1.7.0

PR #270, opened 2021-11-21
Replace TravisCI with Github Actions.

jdm:


jdm:
I'm going to go ahead and merge this; the miri task is failing because of https://github.com/rust-lang/rust/issues/91025, but that would be true on TravisCI as well.

RalfJung:
> I'm going to go ahead and merge this; the miri task is failing because of rust-lang/rust#91025, but that would be true on TravisCI as well.

FWIW, Miri will just occasionally not work for a few nightlies, so the [recommended CI script](https://github.com/rust-lang/miri/#running-miri-on-ci) installs the latest nightly with Miri, instead of relying that the latest nightly has Miri.

PR #271, opened 2021-12-19
Test for drains that shift the tail, when inline

saethlin:
Previously, the test suite only had one trip through the tail-shifting
code in Drain::drop, and that is in the heap state.
In the current implementation, a tail-shifting drain while in the inline
state produces potentially dangerous aliasing which is currently
accepted by default Miri and rejected with -Ztrack-raw-pointers.

Adding this test case ensures that if this ever becomes an actual
problem it will be easy to find.


jdm (approved):

jdm:
@bors-servo r+
Thanks!

saethlin:
The bors failure above was only because the nightly that day didn't have a miri. I was trying to prompt for a retry, but I guess I can't :(

PR #277, opened 2022-02-01
Fix all problems encounted with Miri -Ztag-raw-pointers

saethlin:
I poked at this crate with `-Zmiri-tag-raw-pointers` before, and I was unable to fix what I found (I just added a test case that ruled out one of my wrong ideas https://github.com/servo/rust-smallvec/pull/271). I tried again just now and I guess I just understand better this time.

This PR fixes 3 separate pointer invalidation problems, which are detected by running `MIRIFLAGS=-Zmiri-tag-raw-pointers cargo miri test`.

Depending on how you squint, 2 or 3 of these are https://github.com/rust-lang/unsafe-code-guidelines/issues/133. The last one is _probably_ still present even with late invalidation, because `set_len` does a write through a `&mut`.

It's unclear to me if any of these things that Miri complains about are potentially a miscompilation in rustc due to the use of LLVM `noalias`. But perhaps given how subtle this codebase is overall, it would be best to run the tools on their pickiest settings, even if there are a few things more like a false positive than a real problem.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,6 +61,8 @@ jobs:
       - name: miri
         if: matrix.toolchain == 'nightly'
         run: bash ./scripts/run_miri.sh
+        env:
+          MIRIFLAGS: '-Zmiri-tag-raw-pointers'
 
       - name: fuzz
         if: env.DO_FUZZ == '1'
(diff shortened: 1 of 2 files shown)


mbrubeck (approved):
Thank you!

mbrubeck:
(We need to wait for a nightly with working MIRI, or fix our CI script to find a working nightly build automatically.)

saethlin:
To get Miri building again, it needs to support `readdir64`. I've been meaning to improve Miri's libc support, but don't hold your breath. I definitely do not mind waiting for a nightly where Miri builds.

saethlin:
@mbrubeck Miri is on recent nightlies, I pushed an empty commit (please squash it away) to make CI re-run :partying_face:

PR #280, opened 2022-05-11
The scope of the unsafe block can be appropriately reduced

peamaeq:


mbrubeck:
Thanks for the contribution!  However, I don't feel like this improves maintainability of the code.  I think the current unsafe blocks make better units for auditing invariants, even if they contain a mix of safe and unsafe operations.

PR #282, opened 2022-06-23
Faster insert for the `index == len` case

nnethercote:
This case popped up within rustc, where `insert` was repeatedly used for a `push`-like operation. (But in a context where `push` wasn't always appropriate.) It speeds up rustc on one benchmark by 2% -- not huge, but also not bad for such a small change.

Here is some before and after results for the new benchmark along with the existing `push` benchmarks.

Old:
```
test bench_push                        ... bench:         317 ns/iter (+/- 25)
test bench_push_small                  ... bench:          34 ns/iter (+/- 1)
test bench_insert_push                 ... bench:         627 ns/iter (+/- 32)
test bench_insert_push_small           ... bench:          99 ns/iter (+/- 6)
```
New
```
test bench_push                        ... bench:         295 ns/iter (+/- 21)
test bench_push_small                  ... bench:          36 ns/iter (+/- 3)
test bench_insert_push                 ... bench:         504 ns/iter (+/- 17)
test bench_insert_push_small           ... bench:          70 ns/iter (+/- 4)
```
The first two are testing identical code, and show that the time variation is non-trivial. The latter two show the effect of this PR's code changes, which are well beyond the timing variation.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -96,6 +96,8 @@ make_benches! {
     SmallVec<[u64; VEC_SIZE]> {
         bench_push => gen_push(SPILLED_SIZE as _),
         bench_push_small => gen_push(VEC_SIZE as _),
+        bench_insert_push => gen_insert_push(SPILLED_SIZE as _),
+        bench_insert_push_small => gen_insert_push(VEC_SIZE as _),
         bench_insert => gen_insert(SPILLED_SIZE as _),
         bench_insert_small => gen_insert(VEC_SIZE as _),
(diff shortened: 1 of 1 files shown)


nnethercote:
[This](https://github.com/rust-lang/rust/blob/168a0209002fef203e83989ff641c1b3e1a51859/compiler/rustc_data_structures/src/map_in_place.rs#L93) is the relevant use point within Rust, if you are curious.

mbrubeck (approved):

mbrubeck:
Thanks! Released as v1.8.1: https://crates.io/crates/smallvec/1.8.1

PR #283, opened 2022-06-29
Add retain_mut method (as an alias to retain)

cmyr:
Opening this speculatively, since it's ultimately less work than opening an issue?

I was trying to use this crate as a drop-in replacement for std::Vec, where I am using the retain_mut method. I understand that it isn't a useful distinction here, and I'm not sure if api compatibility is a goal, so if this doesn't make sense for the project I understand. :)


mbrubeck:
Thanks!  Published as version 1.9.0.

PR #286, opened 2022-09-08
Add Natvis definitions and tests for `SmallVec` type

ridwanabdillahi:
This change adds Natvis visualizations for `SmallVec` to help improve the debugging experience on Windows.

Natvis is a framework that can be used to specify how types should be viewed under a supported debugger, such as the Windows debugger (WinDbg) and the Visual Studio debugger.

The Rust compiler does have Natvis support for some types, but this is limited to some of the core libraries and not supported for external crates.

https://github.com/rust-lang/rfcs/pull/3191 proposes adding support for embedding debugging visualizations such as Natvis in a Rust crate. This RFC has been approved, merged and implemented.

This PR adds:

* Natvis visualizations for `SmallVec` with and without the `union` feature.
* Tests for testing visualizers embedded in the `smallvec` crate.
* Updates to the CI pipeline to ensure tests for visualizers are run so they do not break silently.
* Add testing against Windows using the nightly toolchain
* A new debugger_visualizer feature for the `smallvec` crate to enable the unstable debugger_visualizer Rust feature.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,9 +7,8 @@ on:
   workflow_dispatch:
 
 jobs:
-  linux-ci:
-    name: Linux
-    runs-on: ubuntu-latest
+  ci:
+    name: Build/Test
     strategy:
       matrix:
         toolchain: ["stable", "beta", "nightly", "1.36.0"]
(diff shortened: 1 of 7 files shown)


mbrubeck (approved):
Looks good, thanks!

MauriceKayser:
Nice, I was just in need of this! Will this be released as 1.9.1?

PR #287, opened 2022-09-12
Fix broken CI yaml script

ridwanabdillahi:
The previous commit updated the CI yaml script to include the `os` key to the matrix for test jobs but did not set an initial value.

Update the matrix to test on `ubuntu-latest` by default.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,6 +12,7 @@ jobs:
     strategy:
       matrix:
         toolchain: ["stable", "beta", "nightly", "1.36.0"]
+        os: [ubuntu-latest]
         include:
           - toolchain: stable
             env:


ridwanabdillahi:
@mbrubeck This is a fix to a previous change that has just been merged. Can you please take a look? Thanks

mbrubeck (approved):

mbrubeck:
It looks like we need to ship a Cargo.lock if we want our tests to build reproducibly with Rust 1.36.0...

ridwanabdillahi:
> It looks like we need to ship a Cargo.lock if we want our tests to build reproducibly with Rust 1.36.0...

@mbrubeck that seems like it's a separate issue. This change was to unblock the CI from running due to invalid syntax in the yaml script. Should I open an issue for the 1.36.0 test failure?

mbrubeck:
Filed #288 for the Rust 1.36 problem.

PR #289, opened 2022-10-01
Mark Array::size() as #[inline]

nikic:
`Array::size()` is a trivial function -- in fact, it always returns a constant. Mark it as `#[inline]` so it is instantiated in each CGU, and can be optimized without relying on crate-local LTO. See https://github.com/rust-lang/rust/issues/102539 for context.

It looks like SmallVec already marks pretty much all other trivial functions as `#[inline]`, so this seems like an oversight.


mbrubeck (approved):
Thanks!

PR #295, opened 2023-01-11
Add abomonation support

markbt:
The [Abomonation](https://docs.rs/abomonation/0.7.3/abomonation/) crate allows transmutation of types into buffers of memory and back again.  This is useful for caches, where it makes it possible to serve read-only references of the item directly out of the cache memory.

Abomonation implements support for normal vectors, however it needs each type to implement the `Abomonation` trait in order for it to work on them.  We have structures containing smallvecs that we'd like to serialize with Abomonation, but currently can't as it doesn't implement the trait.

This change adds a new feature (`abomonation`) to the smallvec crate which provides abomonation support for `SmallVec` if specified.  The implementation is based on the Abomonation `Vec` implementation, with the unspilled special case accounted for: the data is inline, so doesn't need to be stored separately.

I've put this behind a feature flag as I'm sure most users won't want it.  However we might not be the only ones using both SmallVec and Abomonation, so I thought I'd share it to see if there is interest in incorporating it here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ write = []
 union = []
 specialization = []
 may_dangle = []
+abomonation = ["dep:abomonation"]
 
 # UNSTABLE FEATURES (requires Rust nightly)
 # Enable to use the #[debugger_visualizer] attribute.
@@ -26,6 +27,7 @@ debugger_visualizer = []
 [dependencies]
 serde = { version = "1", optional = true, default-features = false }
 arbitrary = { version = "1", optional = true }
+abomonation = { version = "0.7", optional = true }
 
 [dev_dependencies]
 bincode = "1.0.1"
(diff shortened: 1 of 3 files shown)


markbt:
As an alternative, I've also written https://github.com/TimelyDataflow/abomonation/pull/45 which adds the implementation to the `abomonation` crate via a feature flag there.  I think that might be the preferred option, as it isolates the somewhat crazy implementation to that crate, but I'll leave this PR up for now in case there is interest here.

mbrubeck:
I'm fine either way, but if you think that putting the code in `abomonation` makes more sense, I'll wait and see if that PR is accepted first.

mbrubeck:
(This was auto-closed because the `master` branch was deleted.  You may re-submit this pull request against the `v2` branch.)

PR #297, opened 2023-02-02
Add Bincode 2 support

markbt:
Bincode 2.0 will include its own traits for encoding and decoding, which are more efficient than going via serde.

I opened https://github.com/bincode-org/bincode/pull/612 to add implementations for smallvec, however the recommendation there is that these traits should be implemented in the smallvec crate instead.  This PR adds a new feature `bincode2`, which, when enabled, adds implementations of the new bincode traits for `SmallVec`.

I see a roughly 10% improvement in my benchmarks when using the new bincode traits directly vs using serde, so this would be a useful feature to have.

Because there is already a dev-dependency on Bincode 1 used in tests, I've called this feature `bincode2`.  That can be changed if you'd prefer the feature to be called `bincode`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ debugger_visualizer = []
 [dependencies]
 serde = { version = "1", optional = true, default-features = false }
 arbitrary = { version = "1", optional = true }
+bincode2 = { package = "bincode", version = "2.0.0-rc", optional = true, default-features = false }
 
 [dev_dependencies]
 bincode = "1.0.1"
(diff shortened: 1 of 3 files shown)


mbrubeck:
Thanks!  I would like to wait to publish this until bincode 2.0 has a stable release.  You could also upgrade the dev-dependency to bincode 2 if you like.

mbrubeck:
(This was auto-closed because the `master` branch was deleted.  You may re-submit this pull request against the `v2` branch.)

PR #298, opened 2023-06-03
Use `NonNull` pointer in heap version

AngelicosPhosphoros:
Since `SmallVec` allocate on heap only if capacity is larger than inner capacity and therefore larger than 1, it cannot be null. Also, this is same behaviour as `std::vec::Vec` which always uses non-null pointer, which is dangling if it has zero capacity.

This change reduces size of enum in some cases because it can exploit [niche optimization](https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#niche). For example, `size_of::<SmallVec<[u8; 8]>>` is 24 bytes with changes from this PR while before it was 32 bytes.

Also:
1. Changed some internal APIs to indicate that we work with non-nullable pointers.
2. Updated natvis and fixed bug with incorrect calculation of `is_inline` intrinsic.

### DIFF
--- a/debug_metadata/smallvec.natvis
+++ b/debug_metadata/smallvec.natvis
@@ -1,13 +1,14 @@
 <AutoVisualizer xmlns="http://schemas.microsoft.com/vstudio/debugger/natvis/2010">
   <Type Name="smallvec::SmallVec&lt;array$&lt;*,*&gt;&gt;" Priority="Medium">
-    <Intrinsic Name="is_inline" Expression="$T2 &gt; capacity" />
-    <Intrinsic Name="len" Expression="is_inline() ? capacity : data.variant1.value.__0.__1" />
-    <Intrinsic Name="data_ptr" Expression="is_inline() ? data.variant0.value.__0.value.value : data.variant1.value.__0.__0" />
+    <Intrinsic Name="is_inline" Expression="$T2 &gt;= capacity" />
+    <Intrinsic Name="len" Expression="is_inline() ? capacity : data.variant1.value.len" />
+    <Intrinsic Name="data_ptr" Expression="is_inline() ? data.variant0.value.__0.value.value : data.variant1.value.ptr.pointer" />
 
-    <DisplayString>{{ len={len()} }}</DisplayString>
+    <DisplayString>{{ len={len()} is_inline={is_inline()} }}</DisplayString>
     <Expand>
         <Item Name="[capacity]">is_inline() ? $T2 : capacity</Item>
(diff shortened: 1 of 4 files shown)


AngelicosPhosphoros:
Also, I think this changes make `union` feature useless because it would not provide size benefits anymore.

Another thing is `capacity or size` field in small vec implementation where field acts as length in inline version and as capacity in heap version. I don't see benefits of this compared to storing length directly and storing capacity of allocation in heap version.
On the other hand, storing capacity in heap version allow to store `NonZeroUsize` there and get size of `Option<SmallVec<_>>` smaller when used array is small.

Btw, it seems that CI is broken somehow.

mbrubeck:
Hi, thanks for this work!  Interestingly, I tried doing this before but reverted it in https://github.com/servo/rust-smallvec/commit/0e8f5e90cba6e3f860e392db7daf0cf9d24e0b40 because I couldn’t see any size improvements.  I’m not sure if something was wrong with my implementation, or if I just made a mistake testing it.

The CI errors should be fixed by #299.

AngelicosPhosphoros:
@mbrubeck Hi!
I have a trouble with fixing that CI error because it passes on my PC:
```
running 1 test
test test_debugger_visualizer__cdb ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 41.43s
```

Do you know what can cause the issue?

AngelicosPhosphoros:
@mbrubeck 
I fixed natvis for enum version.

And as I tested, natvis for union version didn't work in first place because windbg cannot read `inline` version for some reason. It writes "expected a member name" in watched variables, for example.

I also added `is_inline` to short debug representation because I think it is useful information.

mbrubeck:
Thanks, this looks good!  I'm going to give it one more careful review before merging it, in the next few days.

mbrubeck (approved):

sffc:
I suspect that this PR increases the code size of `SmallVec::try_reserve`, which is used by a number of entrypoints including `SmallVec::insert_from_slice`. On smallvec 1.10, with link-time optimization, `insert_from_slice` gets inlined to the call site and only ever calls down to `rust_alloc`. However, on smallvec 1.11, the function is no longer inlined and it calls `rust_alloc`, `rust_realloc`, and `rust_dealloc`. This causes about 1 KB more code to be linked into WASM than was previously required when using smallvec. I'm not 100% sure this is the culprit PR but it seems to be the only one between 1.10 and 1.11 touching this code path.

CC @Manishearth @robertbastian

PR #292, opened 2022-12-26
Implemented DrainFilter

L0uisc:
Provides an implementation for DrainFilter (issue #281).

PR #284, opened 2022-08-01
Implementation of RFC #240

sarah-ek:
Implementation of the API described in #240
The API and tests were modified and all the tests pass.
So far the docs aren't written yet. But I wanted to hear the feedback on the current progress before going further.

current benchmark results: there are a few regressions but i believe i can handle them

```
                                                                       master                 this PR       relative difference (lower is better)
test bench_extend                      ... bench:          33 ns/iter (+/- 1)      29 ns/iter (+/- 0)                    -0.121
test bench_extend_from_slice           ... bench:          25 ns/iter (+/- 0)      23 ns/iter (+/- 0)                    -0.079
test bench_extend_from_slice_small     ... bench:           9 ns/iter (+/- 0)       8 ns/iter (+/- 0)                    -0.111
test bench_extend_from_slice_vec       ... bench:          53 ns/iter (+/- 3)      52 ns/iter (+/- 3)                    -0.018
test bench_extend_from_slice_vec_small ... bench:          16 ns/iter (+/- 0)      15 ns/iter (+/- 0)                    -0.062
test bench_extend_small                ... bench:          10 ns/iter (+/- 0)      11 ns/iter (+/- 0)                    +0.100
test bench_extend_vec                  ... bench:          53 ns/iter (+/- 1)      51 ns/iter (+/- 2)                    -0.037
test bench_extend_vec_small            ... bench:          12 ns/iter (+/- 0)      12 ns/iter (+/- 0)                    +0.0
test bench_from_iter                   ... bench:          17 ns/iter (+/- 0)      35 ns/iter (+/- 0)                    +1.058
test bench_from_iter_small             ... bench:           3 ns/iter (+/- 0)      11 ns/iter (+/- 1)                    +2.666
test bench_from_iter_vec               ... bench:          18 ns/iter (+/- 1)      17 ns/iter (+/- 0)                    -0.055
test bench_from_iter_vec_small         ... bench:          12 ns/iter (+/- 0)      12 ns/iter (+/- 0)                    +0.0
test bench_from_slice                  ... bench:          17 ns/iter (+/- 0)      22 ns/iter (+/- 0)                    +0.294
test bench_from_slice_small            ... bench:           3 ns/iter (+/- 0)       7 ns/iter (+/- 0)                    +1.333
test bench_from_slice_vec              ... bench:          18 ns/iter (+/- 0)      17 ns/iter (+/- 0)                    -0.055
[...]

PR #300, opened 2023-06-17
Version 1.11.0

mbrubeck:


jdm (approved):

mbrubeck:
Notable changes include #292 and #298.  Some additional eyes on those changes might be good before we release this.

After this is released, I'd like to make #284 the main branch, and begin preparing for smallvec 2.0.

Manishearth (approved):
r=me for the version bump, haven't looked too closely at the other changes

the8472:
Note that DrainFilter in std has been replaced with [ExtractIf](https://doc.rust-lang.org/nightly/std/vec/struct.ExtractIf.html) which doesn't keep draining on drop.

mbrubeck:
> Note that DrainFilter in std has been replaced with [ExtractIf](https://doc.rust-lang.org/nightly/std/vec/struct.ExtractIf.html) which doesn't keep draining on drop.

Thanks for the heads-up.  We can add this new API behind a separate feature if someone has time to implement it, and we can drop the old API in version 2.0.

PR #305, opened 2023-08-02
ci: Improve CI for v2

waywardmonkeys:
These resulted in warnings in the GitHub Actions UI. The `actions-rs` actions haven't been maintained in some time.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,10 +44,6 @@ jobs:
         if: matrix.toolchain == 'nightly'
         run: cargo check --verbose --no-default-features
 
-      - name: Cargo test w/ union
-        if: matrix.toolchain == 'beta'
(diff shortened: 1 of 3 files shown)


mbrubeck (approved):
Thanks!

waywardmonkeys:
That's odd. There's a `binutils-dev` on Ubuntu jammy, so not sure why this failed. (These packages were installed for running the fuzzer... which I'm not sure is an ideal thing to be doing in CI anyway.)

waywardmonkeys:
(And they installed fine for the checks running on the v1 branch...)

Can you re-trigger the build?

mbrubeck:
#307 will fix the apt errors.

waywardmonkeys:
I'd just realized that as well and added a commit here that fixes that as well.

PR #307, opened 2023-08-02
Fix apt-get errors in CI

mbrubeck:
Forward-port of #299 to the v2 branch.


waywardmonkeys:
I've added the same change to my PR #305 which contains fixes needed to let CI pass at all on the `v2` branch.

mbrubeck:
Fixed by #305.

PR #310, opened 2023-08-03
ci: Fix running fuzz on `v2` branch.

waywardmonkeys:
Running the `fuzz` step wasn't happening, so this is an attempt at fixing that.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,16 +15,14 @@ jobs:
         toolchain: ["stable", "beta", "nightly", "1.57.0"]
         include:
           - toolchain: stable
-            env:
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
This now appears to be actually failing, not just a CI configuration issue.

mbrubeck (approved):

waywardmonkeys:
@mbrubeck While I have the time to fix CI, I don't have the time to debug and figure out why the fuzz is actually failing on `v2`. :(

mbrubeck:
That's fine.  We can leave this open until we can fix the underlying issue.

mbrubeck:
resolved by #332

PR #329, opened 2023-11-18
Add `from_const_with_len_unchecked`

### LINKED ISSUE #301: Add from_const_with_len
What are the projects thoughts on adding `from_const_and_len` version of `from_const` to be able to create a `SmallVec` in a const context that does not logically occupy the full inline stack space? Like:

```rust
#[cfg_attr(docsrs, doc(cfg(feature = "const_new")))]
#[inline]
pub const fn from_const_and_len(items: [T; N], len: usize) -> Self {
    if len >= N {
        panic!("Expected len to be less than or equal to array size");
    }
    SmallVec {
        capacity: len,
        data: SmallVecData::from_const(MaybeUninit::new(items)),
    }
}
```

## Example Application

The reason this came up is that I'm working on a MIDI library. MIDI uses a variable length quantity (VLQ) type for integers. I'm using a `SmallVec<[u8; 4]>` as the backing storage so that a VLQ can store 4 bytes inline before spilling onto the heap. I wanted to be able to create a `VLQ::ZERO` constant to represent `0` stored as a VLQ, which means the backing `SmallVec` needs to be logically `[0]` (capacity = 1). But my only option (as far as I am aware) was `SmallVec::from_const([0; 4])` which would logically be `[0, 0, 0, 0]`, which is not a valid representation of `0` as a VLQ.

## Issues
Unfortunately, [`panic!` in const contexts was only stabilized in Rust 1.57](https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html#panic-in-const-contexts), which as far as I am aware would cause an issue since currently the `const_new` feature only requires Rust 1.51. I'm not sure how (if at all) this can be mitigated to keep `from_const_and_len` safe while not changing the version requirements for the `const_new` feature.

Expyron:
Fixes #301

I made the method `unsafe` to be able to use the existing `const_new` feature without raising its MSRV to 1.57, which would have been necessary for const-panicking.

This is v1-only. v2 already has `const` `from_buf_and_len_unchecked`, so this use-case is already covered there.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2384,6 +2384,20 @@ impl<T, const N: usize> SmallVec<[T; N]> {
             data: SmallVecData::from_const(MaybeUninit::new(items)),
         }
     }
+
+    /// Constructs a new `SmallVec` on the stack from an array without
+    /// copying elements. Also sets the length. The user is responsible
+    /// for ensuring that `len <= N`.
+    /// 
+    /// This is a `const` version of [`SmallVec::from_buf_and_len_unchecked`] that is enabled by the feature `const_new`, with the limitation that it only works for arrays.
+    #[cfg_attr(docsrs, doc(cfg(feature = "const_new")))]
+    #[inline]
+    pub const unsafe fn from_const_with_len_unchecked(items: [T; N], len: usize) -> Self {
+        SmallVec {
+            capacity: len,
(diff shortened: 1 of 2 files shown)


mbrubeck (approved):

mbrubeck:
Sorry for the delay.  This has been published as smallvec 1.12.0.

PR #332, opened 2024-01-13
Correctly handle `Vec` with zero capacity in `from_vec`

matzemathics:
Since the standard library does  not allocate any memory for vectors with capacity 0, the pointer returned by `as_mut_ptr` is dangling in that case. This causes a segfault, when `shrink_to_fit` tries to free.


mbrubeck (approved):
Thanks!

PR #334, opened 2024-01-14
Use NonNull in place of *const (v2)

matzemathics:
Port of #171. While it (at least currently) does not improve the size of objects in any case I could come up with, it still feels a lot more appropriate to use, because it cleanly communicates the requirements it already had: it is covariant and never null.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1080,7 +1080,7 @@ impl<T, const N: usize> SmallVec<T, N> {
                     // so the copy is within bounds of the inline member
                     copy_nonoverlapping(ptr.as_ptr(), self.raw.as_mut_ptr_inline(), len);
                     drop(DropDealloc {
-                        ptr: NonNull::new_unchecked(ptr.as_ptr() as *mut u8),
+                        ptr: NonNull::new_unchecked(ptr.cast().as_ptr()),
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -826,9 +828,9 @@ impl<T, const N: usize> SmallVec<T, N> {
 
                     // SAFETY: len <= new_capacity <= Self::inline_size()
                     // so the copy is within bounds of the inline member
-                    copy_nonoverlapping(ptr, self.raw.as_mut_ptr_inline(), len);
+                    copy_nonoverlapping(ptr.as_ptr(), self.raw.as_mut_ptr_inline(), len);
                     drop(DropDealloc {
-                        ptr: NonNull::new_unchecked(ptr as *mut u8),
+                        ptr: NonNull::new_unchecked(ptr.as_ptr() as *mut u8),
```
mbrubeck: I think this can use [`NonNull::cast`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.cast).

mbrubeck (changes requested):
Thanks again!  Added one minor request.

mbrubeck (approved):

PR #335, opened 2024-01-17
Impl get_size::GetSize (behind feature flag) (v1 branch)

amandasaurus:
Shows memory usage of the struct

cf. issue #331

This includes unittests, and I have briefly used this in real programmes successfully.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -1023,3 +1023,70 @@ fn drain_keep_rest() {
 
     assert_eq!(a, SmallVec::<[i32; 3]>::from_slice(&[1i32, 3, 5, 6, 7, 8]));
 }
+
+#[cfg(all(feature = "get-size", target_pointer_width = "64"))]
+mod get_size {
+    use super::*;
+
+    #[test]
+    fn end_to_end1() {
+        use ::get_size::GetSize;
+
+        let mut a: SmallVec<[i32; 2]> = smallvec![];
+        assert!(!a.spilled());
+        assert_eq!(a.len(), 0);
+        assert_eq!(a.get_size(), 24);
+        assert_eq!(a.get_heap_size(), 0);
+
+        a.push(0);
+        assert_eq!(a.len(), 1);
+        assert!(!a.spilled());
+        assert_eq!(a.get_size(), 24);
+        assert_eq!(a.get_heap_size(), 0);
+
+        a.push(1);
+        assert_eq!(a.len(), 2);
+        assert!(!a.spilled());
+        assert_eq!(a.get_size(), 24);
+        assert_eq!(a.get_heap_size(), 0);
+
+        a.push(2);
+        assert_eq!(a.len(), 3);
+        assert!(a.spilled());
+        assert_eq!(a.get_size(), 40);
+        assert_eq!(a.get_heap_size(), 16);
+
+        a.push(3);
+        assert_eq!(a.len(), 4);
+        assert!(a.spilled());
+        assert_eq!(a.get_size(), 40);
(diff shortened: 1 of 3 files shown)


src/tests.rs
```diff
@@ -1023,3 +1023,51 @@ fn drain_keep_rest() {
 
     assert_eq!(a, SmallVec::<[i32; 3]>::from_slice(&[1i32, 3, 5, 6, 7, 8]));
 }
+
+#[cfg(feature = "get-size")]
```
mbrubeck: It looks like this test is failing when both the `"get-size"` and `"union"` features are enabled.  You can reproduce this failure by running `cargo test --features get-size,union`.

The test will also fail on non-64-bit platforms.  You can fix both issues by changing this line to something like

```rust
#[cfg(all(feature = "get-size", target_pointer_with = "64", not(feature = "union")))]
```

amandasaurus: 👍 Fixed. & new branch (w. clean history) pushed. I've also added some tests for when the `union` feature is active.

mbrubeck (approved):

mbrubeck:
Thanks! Published as smallvec 1.13.0.

amandasaurus:
Unfortunately, this has a bug.

a `Cargo.toml` with:

```toml
[dependencies]
get-size = { version = "0.1.4", features = ["derive"] }
smallvec = { version = "1.13.0", features = ["get-size"] }
```

will fail to compile with:

```
error: cyclic package dependency: package `attribute-derive v0.6.1` depends on itself. Cycle:
package `attribute-derive v0.6.1`
    ... which satisfies dependency `attribute-derive = "^0.6"` (locked to 0.6.1) of package `get-size-derive v0.1.3`
    ... which satisfies dependency `get-size-derive = "^0.1.3"` (locked to 0.1.3) of package `get-size v0.1.4`
    ... which satisfies dependency `get-size = "^0.1"` (locked to 0.1.4) of package `smallvec v1.13.0`
    ... which satisfies dependency `smallvec = "^1.5"` (locked to 1.13.0) of package `proc-macro-utils v0.8.0`
    ... which satisfies dependency `proc-macro-utils = "^0.8.0"` (locked to 0.8.0) of package `attribute-derive-macro v0.6.1`
    ... which satisfies dependency `attribute-derive-macro = "^0.6.1"` (locked to 0.6.1) of package `attribute-derive v0.6.1`
```

However these dependencies works:

```toml
[dependencies]
get-size = { version = "0.1.4", features = ["derive"] }
smallvec = { git = "https://github.com/servo/rust-smallvec.git", rev="v1.13.0", features = ["get-size"] }
```

😖😖 But the `v1.13.0` tag should be the same as the released `1.13.0`, right?

mbrubeck:
Hmm, that's annoying.  As a workaround you can do this:

```toml
[dependencies]
get-size = "0.1.4"
get-size-derive = "0.1"
smallvec = { version = "1.13.0", features = ["get-size"] }
```

and replace `#[derive(GetSize)]` in your code with:

```rust
#[derive(get_size_derive::GetSize)]
```

but things will still break if you have dependencies outside of your control that enable the `"derive"` feature of `get-size`.

Perhaps it would be better to remove these impls from the `smallvec` crate, and add them to the `get-size` crate, if the author is willing.

> But the v1.13.0 tag should be the same as the released 1.13.0, right?

Using a git dependency breaks the cycle because Cargo considers it a different crate than the crates.io dependency, so it ends up building two different copies of the `smallvec` code.

mbrubeck:
I have reverted this change and yanked version 1.13.0 for now, until we figure out the best solution.  Sorry about this. :(

PR #336, opened 2024-01-17
Impl get_size::GetSize (behind feature flag) (v2 branch)

amandasaurus:
Shows memory usage of the struct

cf. issue #331

this is against the `v2` branch. The tests pass, but I haven't tried this code in real uses yet.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -45,6 +45,10 @@
 //! [Rustonomicon](https://doc.rust-lang.org/1.42.0/nomicon/dropck.html#an-escape-hatch).
 //!
 //! Tracking issue: [rust-lang/rust#34761](https://github.com/rust-lang/rust/issues/34761)
+//!
+//! ### `get-size`
+//!
+//! When this optional dependency is enabled, `SmallVec` implements the `get_size::GetSize` trait.
 
 #![no_std]
 #![cfg_attr(docsrs, feature(doc_cfg))]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -2176,3 +2182,22 @@ impl<const N: usize> io::Write for SmallVec<u8, N> {
         Ok(())
     }
 }
+
+#[cfg(feature = "get-size")]
+impl<A: Array> GetSize for SmallVec<A>
```
mbrubeck: The generic types for SmallVec have changed in `v2`.  This should now be `impl<T, const N: usize>`.

src/tests.rs
```diff
@@ -1028,3 +1028,51 @@ fn drain_keep_rest() {
 
     assert_eq!(a, SmallVec::<i32, 3>::from_slice(&[1i32, 3, 5, 6, 7, 8]));
 }
+
+#[cfg(feature = "get-size")]
```
mbrubeck: This test should be disabled for non-64-bit platforms (similar to #335 except the `"union"` feature does not exist in `v2`).

mbrubeck (changes requested):
Looks good, just a couple of minor changes needed.

amandasaurus:
I've fixed those 2 things, and pushed a new branch.

mbrubeck (approved):

PR #337, opened 2024-01-19
Implement `Hash` trait for `SmallVec`

hsanzg:
While calling the `hash` function on a `SmallVec` currently works via dereferencing into `&[T]` (see `test_hash` on `tests.rs`), this is not enough to allow deriving `Hash` on e.g. a struct containing a single field of type `SmallVec`.

I ran into this issue while upgrading one of my crates to the `beta.2` release.


mbrubeck (approved):
Thanks!

PR #338, opened 2024-02-19
Implement `From<[T; M]> for SmallVec<T, N>` for all `M, N` (v2)

zachs18:
(First commit removes the unnecessary `T: Clone` bound from `From<array>` and `From<Vec>` impls that weren't there in v1.)

Allow using `SmallVec::from(array)` and `array.into()` like you can with `Vec`, even when `array.len() != SmallVec::inline_capacity()`. If the source length is longer than the smallvec's inline capacity, it will use heap-allocation, otherwise it will use inline capacity.

<details> <summary> const-eval limitations </summary>

Ideally, the implementation would be as below to explicitly optimize for the three different cases even in debug mode, but it doesn't work because `[T; M]` and `[T; N]` are still considered different types.

```rs
if M > N {
  Self::from_vec(Vec::from(array))
} else if M == N {
  Self::from_buf(array) // error: expected `N`, found `M`
} else {
  // same as PR
}
```

However, with `opt-level >= 1`, Rust appears to reliably compile this PR to either a basically just memcpy (for M <= N) or an alloc and a memcpy (for M > N).

</details>

Backported to v2 as #339

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1909,12 +1909,12 @@ impl<'a, T: Clone, const N: usize> From<&'a [T]> for SmallVec<T, N> {
         slice.iter().cloned().collect()
     }
 }
-impl<T: Clone, const N: usize> From<[T; N]> for SmallVec<T, N> {
+impl<T, const N: usize> From<[T; N]> for SmallVec<T, N> {
     fn from(array: [T; N]) -> Self {
         Self::from_buf(array)
     }
 }
-impl<T: Clone, const N: usize> From<Vec<T>> for SmallVec<T, N> {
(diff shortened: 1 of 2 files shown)


zachs18:
(force-push explanation: I was going to change this to use the same unconditional `with_capacity(m) -> copy_nonoverlapping(..., m)` as I did in the v1 backport, but the compiler was not able to reliably optimize it as aggressively as `if M > N { Self::from(Vec::from(array)) } else ...` at opt-level=1, so I changed it back to that.)

mbrubeck (approved):
Thanks!

PR #339, opened 2024-02-19
Implement `From<A2> for SmallVec<A1>` for all `A2, A1` with the same `Item`  (v1)

### LINKED ISSUE #272: `SmallVec` should implement `From<[T; N]>` (gated by `cfg(feature = "const_generics")`).
The Rust standard library contains these `impl`s:

<details>
<summary>(click to expand search results for <code>From<[...; N]></code>)</summary>

```
$ rg -B1 'From<\[.*; N\]' library/
library/std/src/collections/hash/set.rs
1010-// See also the equivalent impl on HashMap.
1011:impl<T, const N: usize> From<[T; N]> for HashSet<T, RandomState>

library/std/src/collections/hash/map.rs
1174-// See also the equivalent impl on HashSet.
1175:impl<K, V, const N: usize> From<[(K, V); N]> for HashMap<K, V, RandomState>

library/alloc/src/collections/linked_list.rs
1954-#[stable(feature = "std_collections_from_array", since = "1.56.0")]
1955:impl<T, const N: usize> From<[T; N]> for LinkedList<T> {

library/alloc/src/collections/btree/set.rs
1093-#[stable(feature = "std_collections_from_array", since = "1.56.0")]
1094:impl<T: Ord, const N: usize> From<[T; N]> for BTreeSet<T> {

library/alloc/src/collections/binary_heap.rs
1491-#[stable(feature = "std_collections_from_array", since = "1.56.0")]
1492:impl<T: Ord, const N: usize> From<[T; N]> for BinaryHeap<T> {

library/alloc/src/collections/btree/map.rs
2048-#[stable(feature = "std_collections_from_array", since = "1.56.0")]
2049:impl<K: Ord, V, const N: usize> From<[(K, V); N]> for BTreeMap<K, V> {

library/alloc/src/vec/mod.rs
2909-#[stable(feature = "vec_from_array", since = "1.44.0")]
2910:impl<T, const N: usize> From<[T; N]> for Vec<T> {

library/alloc/src/collections/vec_deque/mod.rs
3050-#[stable(feature = "std_collections_from_array", since = "1.56.0")]
3051:impl<T, const N: usize> From<[T; N]> for VecDeque<T> {

library/alloc/src/boxed.rs
1481-#[stable(feature = "box_from_array", since = "1.45.0")]
1482:impl<T, const N: usize> From<[T; N]> for Box<[T]> {
```

</details>

While most are pretty recent (`1.56.0`), the `Vec` one is quite older (`1.44.0`), but either way, `std` has been clearly moving to more by-value array functionality, and I think `SmallVec` should follow this trend.

As a concrete usecase, I'm working on a codebase that uses `SmallVec` throughout, and a lot of them are created with `[...].into_iter().collect()`, which would be nicer to spell as just `[...].into()`.

I already am locked into requiring a Rust version new enough to have stable const generics, and AFAICT the `const_generics` feature of `smallvec` could be used to opt into a new `From` `impl`, but this existing one may be a small issue:
https://github.com/servo/rust-smallvec/blob/218e0bb00548050146adea58f2a96aab32bc4e54/src/lib.rs#L1689

That is, that pre-existing `impl` requires an array as long as the inline capacity of the `SmallVec`, and while it could be swapped out with the more general one with `#[cfg(feature = "const_generics")]`, the extra flexibility could make inference ambiguous (if anyone has relied on it somehow).

Then again, I don't think syntax like `[T; _]` (literally an underscore for the length, to allow it to be inferred) is stable yet, and something like `SmallVec::from(Default::default())` is too ambiguous anyway, so it would take `SmallVec::from(make_array_of_const_generic_len(...))` to be able to cause an issue.

Of course, with a `smallvec` `2.0` (#240), such compatibility issues (however implausible) wouldn't matter, but I'm not sure if that rewrite is coming any time soon.

zachs18:
#338 backported to v1

With `opt-level >= 1`, Rust appears to reliably compile this PR to either a basically just memcpy (for M <= N) or an alloc and a memcpy (for M > N)

Resolves #272

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2016,10 +2016,18 @@ impl<A: Array> From<Vec<A::Item>> for SmallVec<A> {
     }
 }
 
-impl<A: Array> From<A> for SmallVec<A> {
+impl<A1: Array, A2: Array<Item = A1::Item>> From<A2> for SmallVec<A1> {
     #[inline]
-    fn from(array: A) -> SmallVec<A> {
-        SmallVec::from_buf(array)
+    fn from(array: A2) -> SmallVec<A1> {
+        let m = A2::size();
+        let mut this = Self::with_capacity(m);
+        let array = mem::ManuallyDrop::new(array);
+        // SAFETY: m <= this.capacity()
+        unsafe {
+            ptr::copy_nonoverlapping(&*array as *const A2 as *const A2::Item, this.as_mut_ptr(), m);
+            this.set_len(m);
+        }
+        this
     }
 }
 
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -623,6 +623,16 @@ fn test_from() {
     let small_vec: SmallVec<[u8; 128]> = SmallVec::from(array);
     assert_eq!(&*small_vec, vec![99u8; 128].as_slice());
     drop(small_vec);
+
+    let array = [1; 128];
+    let small_vec: SmallVec<[u8; 1]> = SmallVec::from(array);
+    assert_eq!(&*small_vec, vec![1; 128].as_slice());
+    drop(small_vec);
+
+    let array = [99];
+    let small_vec: SmallVec<[u8; 128]> = SmallVec::from(array);
+    assert_eq!(&*small_vec, &[99u8]);
+    drop(small_vec);
 }
 
 #[test]


mbrubeck:
One downside:  The following program used to compile, but now fails because the type of `_v` cannot be inferred:

```
fn main() {
    let _v = smallvec::SmallVec::from([true]);
}
```

This is still considered a [minor change](https://rust-lang.github.io/rfcs/1105-api-evolution.html#minor-change-generalizing-to-generics-1) by Rust’s versioning conventions, but if it’s likely to break code in the wild then I might prefer to keep this as a v2-only change.

Using `SmallVec::from([a, b, c])` seems like a plausible way to construct a SmallVec, but I don’t know if it’s at all common in practice.

mbrubeck:
I'm going to be conservative and close this PR, to avoid breaking any v1 users.  The new functionality will be available in v2.  Thanks again for the contribution!

PR #340, opened 2024-03-03
 Implement SmallVec::split_off 

arthurprs:
Implements `SmallVec::split_off`, which is like an optimized `from_iter(other.drain(at..))`. This also includes some recent developments from [Vec](https://github.com/rust-lang/rust/pull/119917).

Also includes a commit for mentioning `drain_filter` feature in the top level docs.


mbrubeck (approved):
Thanks!

PR #345, opened 2024-03-20
Fix UB on out-of-bounds insert()

### LINKED ISSUE #343: `SmallVec::insert` is UB on large arguments
This example fails in miri, you can try it on the [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9b3d536d340a9c4e42e487a4ad3d0cac):
```rust
fn main() {
    use smallvec::{smallvec, SmallVec};

    // This SmallVec can hold up to 4 items on the stack:
    let mut v: SmallVec<[i32; 4]> = smallvec![1, 2, 3, 4];

    // It will automatically move its contents to the heap if
    // contains more than four items:
    v.push(5);

    // SmallVec points to a slice, so you can use normal slice
    // indexing and other methods to access its contents:
    v[0] = v[1] + v[2];
    v.sort();
    v.insert(10, 6);
}
```

mbrubeck:
Fixes #343.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1372,13 +1372,14 @@ impl<A: Array> SmallVec<A> {
             }
             let mut ptr = ptr.as_ptr();
             let len = *len_ptr;
+            if index > len {
+                panic!("index exceeds length");
+            }
+            // SAFETY: add is UB if index > len, but we panicked first
             ptr = ptr.add(index);
             if index < len {
+                // Shift element to the right of `index`.
                 ptr::copy(ptr, ptr.add(1), len - index);
-            } else if index == len {
-                // No elements need shifting.
-            } else {
-                panic!("index exceeds length");
             }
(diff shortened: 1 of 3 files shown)


mrobinson (approved):

src/lib.rs
```diff
@@ -1372,13 +1372,13 @@ impl<A: Array> SmallVec<A> {
             }
             let mut ptr = ptr.as_ptr();
             let len = *len_ptr;
+            if index > len {
+                panic!("index exceeds length");
+            }
```
workingjubilee, line 1377: This panic placement is load-bearing and probably should have a SAFETY comment:

```suggestion
            }
            // SAFETY: add is UB if index > len, but we panicked first
```

workingjubilee:
The nightly workflow is fixed by https://github.com/servo/rust-smallvec/pull/346

PR #346, opened 2024-03-20
Add more tests for UB

workingjubilee:
Essentially motivated by https://github.com/servo/rust-smallvec/issues/343 in effect.

Also fixes the miri workflow.


mbrubeck (approved):
Thank you!

PR #347, opened 2024-03-23
bugfix for append not set self.len properly

alanlzhang:
`append` currently not properly `set_len` for self.

This patch fix it.


mbrubeck (approved):
Thank you!

PR #352, opened 2024-05-10
[v1] Make test_size test aware of target pointer width and "union" feature.

### LINKED ISSUE #330: test_size fails on 32-bit.
Debian CI has noticed that test_size in smallvec 1.x (the test does not seem to exist on smallvec 2.x) fails on 32-bit.

With "union" feature enabled.

    ---- tests::test_size stdout ----
    thread 'tests::test_size' panicked at 'assertion failed: `(left == right)`
      left: `24`,
     right: `12`', src/tests.rs:990:5

Without "union" feature enabled.

    ---- tests::test_size stdout ----
    thread 'tests::test_size' panicked at 'assertion failed: `(left == right)`
       left: `24`,
      right: `16`', src/tests.rs:990:5

I applied a patch in Debian to make 24 a maximum value rather than an exact value.

https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/smallvec/debian/patches/fix-test-32-bit.patch

zachs18:
Fixes #330.

Tested locally on `x86-64-unknown-linux-gnu` and `i686-unknown-linux-gnu` with and without `--features union`.

(This does assume that `alignof(usize) == sizeof(usize) == alignof(ptr) == sizeof(ptr)`.)

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -996,7 +996,23 @@ fn test_clone_from() {
 #[test]
 fn test_size() {
     use core::mem::size_of;
-    assert_eq!(24, size_of::<SmallVec<[u8; 8]>>());
+    const PTR_SIZE: usize = size_of::<usize>();
+    #[cfg(feature = "union")]
+    {
+        assert_eq!(3 * PTR_SIZE, size_of::<SmallVec<[u8; 0]>>());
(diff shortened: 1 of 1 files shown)


mbrubeck (approved):
Thanks!

Issue #353, opened 2024-05-10
Segfault on iter().collect() (2.0.0-alpha-5)


josephg:
I'm running into a segfault when calling `some_iter.collect()` into a SmallVec. The problem happens when the smallvec spills and needs to allocate, eventually leading to a segfault in the realloc call.

I'll try and make a minimal reproducing test.

The problem shows up in both debug and release mode, running on the latest stable rust compiler (1.78.0).

Smallvec 1.x works fine.

Valgrind has the gory details. The first problem it spots is an invalid memmove:

```
==21584== Thread 2 list::op_iter:::
==21584== Invalid write of size 8
==21584==    at 0x485293B: memmove (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==21584==    by 0x3F41C2: write<diamond_types::list::operation::TextOperation> (mod.rs:1493)
==21584==    by 0x3F41C2: write<diamond_types::list::operation::TextOperation> (mut_ptr.rs:1491)
==21584==    by 0x3F41C2: smallvec::SmallVec<T,_>::push_heap (lib.rs:968)
==21584==    by 0x3EA6EF: smallvec::SmallVec<T,_>::extend_impl (lib.rs:1519)
==21584==    by 0x3E481F: <smallvec::SmallVec<T,_> as core::iter::traits::collect::FromIterator<T>>::from_iter (lib.rs:1838)
==21584==    by 0x291CA3: core::iter::traits::iterator::Iterator::collect (iterator.rs:2003)
```

Full valgrind trace below.
<details>
```
$ valgrind target/debug/deps/diamond_types-f42d3b862ee83a0c --ignored fix_cs --nocapture
==21584== Memcheck, a memory error detector
==21584== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==21584== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==21584== Command: target/debug/deps/diamond_types-f42d3b862ee83a0c --ignored fix_cs --nocapture
==21584== 

running 1 test
==21584== Thread 2 list::op_iter:::
==21584== Invalid write of size 8
==21584==    at 0x485293B: memmove (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==21584==    by 0x3F41C2: write<diamond_types::list::operation::TextOperation> (mod.rs:1493)
[...]

PR #354, opened 2024-05-10
Added test for servo#353 and fixed use after free bug that causes it

josephg:
I took a look at what was causing #353.

Turns out the `push_heap` function was placing the heap pointer into a local variable before calling `self.reserve()`. reserve in turn calls `realloc()` - which can free the old heap allocation & create a new one. When this happens, the old `ptr` is becomes invalid. 

The code was writing to this invalid pointer (`ptr.as_ptr().add(len).write(value)`) - which is a use-after-free bug.

This PR fixes the bug and adds a regression test.


mbrubeck (approved):
Thanks!

PR #356, opened 2024-05-13
feat: replace `write` feature with `std` feature

### LINKED ISSUE #355: Implement `std::error::Error` for `CollectionAllocErr`
# Description

It seems unnatural that `CollectionAllocErr` currently does not implement `std::error::Error` which would come in handy in cases such as implementing `std::error::Error::source` on user-defined errors caused by `CollectionAllocErr`.

While `error_in_core` is not stable yet, such implementation could be hidden behind a feature flag such as (commmonly used) `std`.

Since the features should be additive (i.e. an inverted `no-std` feature would not be the  right approach), it is a breaking change to add such a feature so I suggest adding this to #183.

If this change is welcome, I may create a pull request to implement it.

cc @mbrubeck as the maintainer of the `2.0` wishlist

JarvisCraft:
# Description

This adds a new `std` feature to enable `std`-specific trait implementations.
This also implements `std::error::Error` on `CollectionAllocErr`.

This resolves #355

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 documentation = "https://docs.rs/smallvec/"
 
 [features]
-write = []
+std = []
 specialization = []
 may_dangle = []
 extract_if = []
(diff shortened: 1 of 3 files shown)


mbrubeck (approved):
Thanks!

PR #359, opened 2024-07-30
[v2] Update to Rust 2021 Edition.

zachs18:
`smallvec` v2's MSRV is currently 1.57.0, which is after the release of Edition 2021 in Rust 1.56.0.

(Also applies a change to `Cargo.toml` changing the deprecated `dev_dependencies` to the preferred `dev-dependencies`, which `cargo fix --edition` warned about but won't actually be an error until Edition _2024_.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "smallvec"
 version = "2.0.0-alpha.7"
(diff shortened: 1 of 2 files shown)


mbrubeck (approved):
Thanks!

PR #364, opened 2024-12-10
Implement spare_capacity_mut

DaniPopes:
Mirrors the implementation from `Vec::spare_capacity_mut`.

This is useful to write directly into the buffer after `SmallVec::with_capacity`, removing the overhead of the `SmallVec` model entirely when doing so.

Should I open a PR for v1 as well?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1413,6 +1413,22 @@ impl<T, const N: usize> SmallVec<T, N> {
         }
     }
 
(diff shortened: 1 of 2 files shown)


mbrubeck (approved):
Thanks!

mbrubeck:
> Should I open a PR for v1 as well?

You don’t need to, but you may do this if you want to use this in a project that depends on smallvec v1.

PR #369, opened 2025-02-15
Implement `MallocSizeOf` for SmallVec (v2)

nicoburns:
Now that [malloc_size_of](https://github.com/servo/malloc_size_of) is published to crates.io we are moving the trait implementations of [MallocSizeOf](https://docs.rs/malloc_size_of/latest/malloc_size_of/trait.MallocSizeOf.html) and related traits into the crates that define the types. This will avoid the situation where depending on `malloc_size_of` pulls in a large number of dependencies.

This PR adds the implementation of `MallocSizeOf` to the 2.x version of the `smallvec` crate.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,9 @@ jobs:
       - name: Cargo test w/ serde
         run: cargo test --verbose --features serde
 
+      - name: Cargo test w/ malloc_size_of
+        run: cargo test --verbose --features malloc_size_of
+
       - name: Cargo check w/o default features
         if: matrix.toolchain == 'nightly'
         run: cargo check --verbose --no-default-features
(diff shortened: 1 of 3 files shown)


nicoburns:
Not sure why this is failing CI it's complaining about `malloc_size_of` not existing on crates.io. But it definitely does (https://crates.io/crates/malloc_size_of), and the other CI jobs are finding it. Some kind of caching issue?

mbrubeck (approved):

nicoburns:
Turns out this was failing due to using Cargo features (`dep:` syntax) that were unsupported in Cargo 1.57. This has been resolved by removing the use of those features in `malloc_size_of`.

mbrubeck (approved):

PR #370, opened 2025-02-15
Implement `MallocSizeOf` for SmallVec (v1)

nicoburns:
Now that [malloc_size_of](https://github.com/servo/malloc_size_of) is published to crates.io we are moving the trait implementations of [MallocSizeOf](https://docs.rs/malloc_size_of/latest/malloc_size_of/trait.MallocSizeOf.html) and related traits into the crates that define the types. This will avoid the situation where depending on `malloc_size_of` pulls in a large number of dependencies.

This PR adds the implementation of `MallocSizeOf` to the 1.x version of the `smallvec` crate.


mbrubeck (approved):

mbrubeck:
Published in smallvec 1.14.0.

nicoburns:
@mbrubeck Just wanted to say thanks for the quick followup on these - very much appreciated!

PR #371, opened 2025-03-20
feat: impl `bytes::BufMut` for `SmallVec` (v2)

BugenZhao:
The [`bytes`](https://crates.io/crates/bytes) crate is a popular crate for manipulating bytes efficiently. Implementing the `BufMut` trait for `SmallVec` would allow it to serve as a seamless replacement for `Vec<u8>` when working with `bytes`.


mbrubeck (approved):
Thank you!

PR #374, opened 2025-03-21
Fix `clippy::legacy_numeric_constants` lints

waywardmonkeys:


nicoburns (approved):

nicoburns (changes requested):
This seems to be failing MSRV checks. smallvec had a very conservative MSRV.

waywardmonkeys:
Ah, never mind then.

Issue #183, opened 2019-11-19
[meta] Wishlist for smallvec 2.0


mbrubeck:
These are changes that weren't ready in time for 1.0 (#73).

Forward port changes from the v1 branch:

- [X] drop bincode2 support #426 

Requires newer version of Rust than current minimum (1.36):

- [x] Use const generics to change `SmallVec<[T; N]>` to `SmallVec<T, N>`. Remove the `Array` trait. (#240, Rust 1.51).
- [x] Use the `union` representation by default and remove the optional `union` feature added in #94 (Rust 1.49).
- [x] Make `SmallVec::new` a `const fn` (#263).

Depends on unstable Rust features:

- [ ] Support custom allocators (#55, rust-lang/rust#32838).
- [x] Use specialization to optimize `extend`, `insert_many`, and `from` for slices of `Copy` types. Deprecate and remove `extend_from_slice`, `insert_from_slice`, and `from_slice` (#384, rust-lang/rust#31844).
- [x] Implement the `Error` trait without requiring `std` (#355, rust-lang/rust#103765).

Breaking API changes:

- [x] Fix variance of `SmallVec<[T; N]>` over `T` (#146)?
- [x] Remove the deprecated `ExtendFromSlice` trait.
- [x] Remove unused `&self` parameter from `inline_size` method.
- [x] Deprecate the `write` feature and add a `std` feature instead (#355, #356).
- [x] Remove the non-standard `insert_many` method (#255)?

protodeniz:
`std::io::write` will move into `core` most likely, according to https://github.com/rust-lang/rust/issues/154046. maybe it would be nice to squash that in here as well and ungate (`std`) the implementation of `Write` trait?

alejandro-vaz:
@protodeniz agree, but it seems that it is taking a while

hopefully it stabilizes before the end of the year

alejandro-vaz:
the wishlist is basically fully implemented, except for allocators, but given that it might be stabilizing soon that might be a better option

@TDecking has been working on it, he knows better than me

the current wishlist is our issues tab, so this is not of much use anymore

protodeniz:
> [@protodeniz](https://github.com/protodeniz) agree, but it seems that it is taking a while
> 
> hopefully it stabilizes before the end of the year

[...]

PR #375, opened 2025-04-04
Implement bincode2 encode/decode support for smallvec v1

markbt:
We would like to use [bincode v2](https://github.com/bincode-org/bincode) to encode some structures that contain smallvecs.  Now that bincode v2 has been released, we can add optional implementations of its traits to smallvec.

This PR implements support in smallvec v1 if the optional `impl_bincode` feature is specified.  Similarly to how bincode has optimizations for `Vec<u8>`, it includes optimizations for `SmallVec<[u8; N]>`.

Note that I have not updated the existing dev-dependency on `bincode` that is used in tests.  This is because bincode v2 has a fairly high MSRV (1.85.0).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ specialization = []
 may_dangle = []
 drain_filter = []
 drain_keep_rest = ["drain_filter"]
+impl_bincode = ["bincode", "unty"]
 
 # UNSTABLE FEATURES (requires Rust nightly)
 # Enable to use the #[debugger_visualizer] attribute.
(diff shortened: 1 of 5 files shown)


mbrubeck (changes requested):
Thanks, this looks good!

Cargo.toml
```diff
@@ -20,6 +20,7 @@ specialization = []
 may_dangle = []
 drain_filter = []
 drain_keep_rest = ["drain_filter"]
+impl_bincode = ["bincode", "unty"]
```
mbrubeck, line 23: Please change the feature name to `bincode` and use `dep:` syntax:

```toml
bincode = ["dep:bincode", "dep:unty"]
```

markbt: Happy to do this, but I went with this approach as `Cargo.toml` says the MSRV is 1.57 and "dep:" syntax was introduced in [1.60](https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#new-syntax-for-cargo-features).

mbrubeck: Got it.  No change needed, then.

mbrubeck (approved):

bgw:
Does this need to be ported to the v2 branch?

mbrubeck:
> Does this need to be ported to the v2 branch?

Yes.  Added to the checklist in #183.

PR #377, opened 2025-04-27
Constant new SmallVec<T, N> for arrays smaller than N

otcova:
## What is this about
We should be able to create a constant SmallVec<T, N> of any size from 0 to N.
Currently, we have the following:
```rust
// Ok since N = 4
const DATA: SmallVec<i32, 4> = SmallVec::from_buf([1, 2, 3, 4]);

// Error
const DATA: SmallVec<i32, 4> = SmallVec::from_buf([1, 2, 3]);
```

## This PR
I've implemented a new `from_buf` that takes an array of size [T; S] instead of [T; N].

## Does it conflict with the current `from_buf`?
If we move it to a different function and keep the current `from_buf` how it is now, it's fine for me.

## Performance concerns
I've copied the code into https://godbolt.org to check that the compiler is able to optimize the copy into a new buffer.
Here are the results:
```rust
use std::ptr::copy_nonoverlapping;
use std::mem::MaybeUninit;

#[inline]
pub const fn from_buf<T, const S: usize, const N: usize>(elements: [T; S]) -> MaybeUninit<[T; N]> {
    assert!(S <= N);
    let mut buf: MaybeUninit<[T; N]> = MaybeUninit::uninit();
    unsafe {
        copy_nonoverlapping(elements.as_ptr(), buf.as_mut_ptr() as *mut T, S);
    }
    std::mem::forget(elements);
    buf
}

#[inline(never)]
pub fn S_eq_N(num: [i32; 4]) -> MaybeUninit<[i32; 4]>  {
    from_buf(num)
}

#[inline(never)]
pub fn S_eq_N_ideal(num: [i32; 4]) -> MaybeUninit<[i32; 4]>  {
    MaybeUninit::new(num)
}

#[inline(never)]
pub fn S_less_than_N(num: [i32; 3]) -> MaybeUninit<[i32; 4]>  {
    from_buf(num)
}

#[inline(never)]
pub fn S_less_than_N_ideal(num: [i32; 3]) -> MaybeUninit<[i32; 4]>  {
    MaybeUninit::new([num[0], num[1], num[2], unsafe { MaybeUninit::uninit().assume_init() }])
}
```
Assembly with `-C opt-level=1`:
```asm
example::S_eq_N::h49a2fa36b6c32877:
        mov     rax, rdi
        movups  xmm0, xmmword ptr [rsi]
        movups  xmmword ptr [rdi], xmm0
        ret

example::S_eq_N_ideal::hfc5842e0def5e8e0:
        mov     rax, rdi
        movups  xmm0, xmmword ptr [rsi]
        movups  xmmword ptr [rdi], xmm0
        ret

example::S_less_than_N::hab2b36d48e2ebe64:
        mov     rax, rdi
        mov     ecx, dword ptr [rsi + 8]
        mov     dword ptr [rdi + 8], ecx
        mov     rcx, qword ptr [rsi]
        mov     qword ptr [rdi], rcx
        ret

example::S_less_than_N_ideal::he3468960a62a802e:
        mov     rax, rdi
        mov     rcx, qword ptr [rsi]
        mov     edx, dword ptr [rsi + 8]
        mov     qword ptr [rdi], rcx
        mov     dword ptr [rdi + 8], edx
        ret
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -673,11 +673,27 @@ impl<T, const N: usize> SmallVec<T, N> {
     }
 
     #[inline]
-    pub const fn from_buf(buf: [T; N]) -> Self {
-        // SAFETY: all the members in 0..N are initialized
+    pub const fn from_buf<const S: usize>(elements: [T; S]) -> Self {
+        assert!(S <= N); // Free check since the values are known at compile time
+
+        // Althought we create a new buffer, since S and N are known at compile time,
+        // even with `-C opt-level=1`, it gets optimized as best as it could be. (Checked with <godbolt.org>)
+        let mut buf: MaybeUninit<[T; N]> = MaybeUninit::uninit();
+
+        // SAFETY: buf and elements do not overlap, are aligned and have space
+        // for at least S elements since S <= N.
+        // We will drop the elements only once since we do forget(elements).
+        unsafe {
+            copy_nonoverlapping(elements.as_ptr(), buf.as_mut_ptr() as *mut T, S);
+        }
+
+        // `elements` have been moved into buf and will be droped by SmallVec
+        core::mem::forget(elements);
+
+        // SAFETY: all the members in 0..S are initialized
         Self {
-            len: TaggedLen::new(N, false, Self::is_zst()),
-            raw: RawSmallVec::new_inline(MaybeUninit::new(buf)),
+            len: TaggedLen::new(S, false, Self::is_zst()),
+            raw: RawSmallVec::new_inline(buf),
             _marker: PhantomData,
         }
     }


mbrubeck (approved):
Thanks!

mbrubeck:
This fails to build in Rust 1.57, so we will need to bump our MSRV if we want to ship this.  It looks like it might require Rust 1.83? 

(Also, I'm not sure why this made it through the merge queue without passing all of the tests.  Looks like we need to fix our merge queue configuration.)

nsabovic:
@otcova thank you so much for this PR. It made our code 5-10% faster on a bunch of benchmarks because we frequently create SmallVec with one element and the compiler is able to make much better code with this.

PR #380, opened 2025-05-27
Unconditionally implement `core::error::Error`

TDecking:
The `error_in_core` feature is stable since Rust 1.81, below this crate's MSRV.


mbrubeck (approved):
Thanks!

PR #381, opened 2025-05-27
Update the `extract_if` feature

TDecking:
This will bring the `extract_if` feature to the same level as currently in libstd. Rust 1.87 actually stabilized its version of `extract_if`, but that is a version above this crate's MSRV, so the feature itself has not been touched.


mbrubeck (approved):
Thanks again!

PR #382, opened 2025-05-29
Add missing `Vec` functionality

### LINKED ISSUE #255: Deprecate insert_many?
The `insert_many` function is complicated, and has had several bugs and vulnerabilities (#96, #208, #252).  Unlike most `SmallVec` methods, it does not correspond to a standard `Vec` method.  Perhaps because of this, it is rarely used.  Should we deprecate this method, and remove it in the next major version?

The standard `Vec::splice` method could be used in place of `insert_many`, but `SmallVec` does not yet implement this method.  We could add this, possibly by copying the implementation from `std::vec`, if there is demand for it.

TDecking:
This will add missing `const` qualifiers, various stable functions, and some performance optimizations from `Vec` to `SmallVec`.
Among them is `splice`, which can replace uses of `insert_many`. The latter has been removed, which should fix #255.

As a small bit of refactoring, functions that rely exclusively on the global allocator (such as `with_capacity`) have been put in a separate implementation block. That will make it a bit easier to add an allocator api later on.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -279,25 +279,6 @@ fn gen_from_elem<V: Vector<u64>>(n: usize, b: &mut Bencher) {
 }
 
 #[bench]
-fn bench_insert_many(b: &mut Bencher) {
-    #[inline(never)]
-    fn insert_many_noinline<I: IntoIterator<Item = u64>>(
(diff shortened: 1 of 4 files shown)


mbrubeck (approved):
Thanks! This is much appreciated.

PR #384, opened 2025-06-02
Rework specialization

TDecking:
This ports most of the specialization work done for the `Vec` Rust type to this library. With that the `extend`, `from`, and `extend_from_slice` are now specialized, with specialization triggering in more instances. This also extends to other functions such as `from_elem`  and `clone_from` which were not specialized before.

This also adds some missing trait implementations and functions with trait bounds.

All of this should finish one of the items in #183. Note that `extend_from_slice` has _not_ been removed, since `Vec`
also has this function. Instead, its trait bound was relaxed from `Copy` to `Clone` for a matching signature.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -80,7 +80,7 @@ impl<T: Copy> Vector<T> for SmallVec<T, VEC_SIZE> {
     }
 
     fn from_elems(val: &[T]) -> Self {
-        SmallVec::from_slice(val)
+        SmallVec::from(val)
     }
 
     fn extend_from_slice(&mut self, other: &[T]) {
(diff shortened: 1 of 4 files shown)


mbrubeck:
I'm not sure we should make the exact same API decisions as the standard library here.

Because of the standard library's special status, it can use specialization internally even with the stable toolchain.  However, smallvec cannot do the same.  We should expect (and recommend) that most users of this crate will not have the `specialization` feature enabled.  But they still may need access to the specially-optimized versions of these functions when applicable.

Therefore, I think we might want to continue offering optimized `Copy`-only functions like the current `from_slice` and `extend_from_slice` (though possibly not under the same name, since we should also provide the less-specialized version that matches the `std::vec` name and signature).

TDecking:
@mbrubeck I've added some `Copy`-only functions back. They have a `_copy` suffix in order to distinguish them from the others.

mbrubeck:
Thanks.  I will try to finish reviewing this soon.

TDecking:
@mbrubeck Any update?

mbrubeck (approved):
Sorry for the delay, and thanks again for this work!

PR #385, opened 2025-06-04
Shrink code produced by `smallvec![]`.

nnethercote:
Currently `smallvec![]` expands to this:
```
{
    let count = 0usize;
    #[allow(unused_mut)]
    let mut vec = ::smallvec::SmallVec::new();
    if count <= vec.capacity() {
	vec
    } else {
	::smallvec::SmallVec::from_vec(::alloc::vec::Vec::new())
    }
};
```
This commit adds a rule to the `smallvec!` macro for the zero-length case so it instead expands to this:
```
::smallvec::SmallVec::new()
```
The `std::vec!` macro already has a similar special case.

This commit also improves the non-zero case.
- It removes the `#[allow(unused_mut)]`, which was only needed for the zero-length case.
- It changes the `*` repetitions to `+`. (Again, like `std::vec`.)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2140,18 +2140,20 @@ impl<T: Clone, const N: usize> Clone for IntoIter<T, N> {
 macro_rules! smallvec {
     // count helper: transform any expression into 1
     (@one $x:expr) => (1usize);
+    () => (
+        $crate::SmallVec::new()
+    );
     ($elem:expr; $n:expr) => ({
         $crate::SmallVec::from_elem($elem, $n)
     });
-    ($($x:expr),*$(,)?) => ({
-        let count = 0usize $(+ $crate::smallvec!(@one $x))*;
-        #[allow(unused_mut)]
+    ($($x:expr),+$(,)?) => ({
+        let count = 0usize $(+ $crate::smallvec!(@one $x))+;
         let mut vec = $crate::SmallVec::new();
         if count <= vec.capacity() {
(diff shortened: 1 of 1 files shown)


nnethercote:
This helps with compile times. I wrote a stress test with 5,000 `smallvec![]` calls, like this:
```rust

fn main() {
    type S = SmallVec<u32, 2>;
    let _: S = smallvec![];
    let _: S = smallvec![];
    let _: S = smallvec![];
    // ... 4997 more repetitions ...
}
```

And the compile time improved dramatically:

|     |old  |new |
|-----|-----|----|
|check| 9.6s|0.5s|
|build|11.3s|0.8s|

mbrubeck (approved):
Thanks!

nnethercote:
@mbrubeck: is it worth backporting this to 1.x? I'm not sure what the status of 1.x is, whether future releases are likely, etc.

mbrubeck:
Yes, I think this is worth backporting.  1.x will probably be used more than 2.x for quite a while yet.

nnethercote:
Ok. What's the process? Is it something you can do, or should I do it? If the latter, what do I do? Just make a PR against the v1 branch? I've never made a PR against the non-default branch but I guess there's a way to do it.

nicoburns:
> I've never made a PR against the non-default branch but I guess there's a way to do it.

You change the "base" in the PR creation UI:

Screenshot 2025-06-06 at 00 28 21

PR #388, opened 2025-07-13
feat: `litemap` store backend

lightsing:
This pull request implements the store backend of [`litemap`](https://docs.rs/litemap/0.8/litemap/), a highly simplistic “flat” key-value map based off of a single sorted vector.

This feature is gated behind `litemap` feature.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,6 +38,9 @@ jobs:
       - name: Cargo test
         run: cargo test --verbose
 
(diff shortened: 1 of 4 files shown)


mbrubeck:
Sorry for the delay in reviewing this.  Just wanted to let you know this is still on my to-do list.

alejandro-vaz:
we should not implement this

we'll make the internals public instead behind a feature gate so you can make your own implementation

see #348 

thanks for contributing, though

Issue #389, opened 2025-09-02
Add defmt support


darkwater:
SmallVec is useful for embedded applications, and [defmt](https://defmt.ferrous-systems.com/) is a popular logging framework for embedded. It would be useful to add a `defmt` feature that implements `defmt::Format` for all types.

alejandro-vaz:
only implement this into v2 since v1 API is locked and only accepts PRs for maintenance

PR #390, opened 2025-09-10
Implement defmt for v1

darkwater:
Implements #389 for the v1 branch


alejandro-vaz:
v1 branch has its API locked, the implementation will go on v2

it must pass tests before

Issue #376, opened 2025-04-26
Add `encase` trait implementations


teoxoy:
I would like to remove trait implementations on foreign types from encase for the reasons listed in https://github.com/teoxoy/encase/issues/94.

They are currently hosted here: https://github.com/teoxoy/encase/blob/fcc83d706f4a51bfbe22c9901973f1c2fa806526/src/impls/smallvec.rs

Given that I could only find [one project](https://github.com/huggingface/ratchet/blob/136da4d5216910bfd015b27a17b837c21f17163a/crates/ratchet-core/Cargo.toml#L32) (and [one if its hard? forks](https://github.com/vinhowe/piston/blob/db808db692f77097b75dabffe56f2a27a050c96b/crates/ratchet-core/Cargo.toml#L33)) using the smallvec feature of encase, would the smallvec developers be willing to host the trait implementations?

mbrubeck:
Yes, we could add the impls to this crate.

alejandro-vaz:
@teoxoy would you be able to implement if for v2??

bolshoytoster:
If not, I can implement it.

alejandro-vaz:
yeah go ahead implement it

teoxoy:
If you could based on https://github.com/servo/rust-smallvec/pull/392 that would be great! Thanks!

PR #392, opened 2025-09-12
Add `encase` trait implementations

teoxoy:
Resolves https://github.com/servo/rust-smallvec/issues/376.


alejandro-vaz:
v1 API is locked down, it only accepts bug fixes

will consider it for v2

PR #393, opened 2025-09-14
Switch `serde` dependency to `serde_core` for v2

paolobarbolini:
[serde_core](https://crates.io/crates/serde_core) has just been released. Crates that don't depend on serde derive macros should use it instead of `serde`, as it speeds up compile times by having the build for the crate itself be independent from `serde-derive` (in case it were to be enabled by some other crate). See the serde_core docs for more info.

On my Ryzen 5 5500U, the clean release build time for just `smallvec` in a project containing `smallvec` with the `serde` feature enabled, and `serde` with the `derive` feature enabled, went from 4.61s to 2.34s.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,10 +17,11 @@ std = []
 specialization = []
 may_dangle = []
 extract_if = []
+serde = ["dep:serde_core"]
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


nicoburns (approved):
I can't think of any reason why this wouldn't be fine.

PR #396, opened 2025-12-08
Refactor v2: faster pop/reserve/retain_mut, modern Criterion benches, Miri-friendly tests + fmt in CI

fereidani:
Hey,
I did some improvements for the library.

What’s in here:
- Switched the benchmarks over to Criterion and used `black_box` for correct more accurate benchmark and added `retain_mut` benches so the `Vector` trait is fully covered now.
- Made the tests less time consuming when running under Miri.
- Reworked `retain_mut`, `reserve`/`reserve_exact`, and `pop`. They’re cleaner now and faster.
- Added a fmt check to CI to keep the library formatted.

here is the benchmark result for retain function:
```
bench_retain_mut_half   time:   [76.825 ns 76.946 ns 77.077 ns]
                        change: [−15.591% −15.307% −15.010%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe

bench_retain_mut_half_small
                        time:   [17.323 ns 17.376 ns 17.434 ns]
                        change: [−5.3509% −4.9664% −4.5552%] (p = 0.00 < 0.05)
                        Performance has improved.

bench_retain_mut_all    time:   [26.311 ns 26.383 ns 26.457 ns]
                        change: [−2.3774% −1.9115% −1.4710%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe

bench_retain_mut_all_small
                        time:   [17.984 ns 18.211 ns 18.453 ns]
                        change: [+0.7628% +2.1217% +3.4929%] (p = 0.00 < 0.05)
                        Change within noise threshold.

bench_retain_mut_none   time:   [25.847 ns 25.917 ns 25.994 ns]
                        change: [−1.2565% −0.3335% +0.5576%] (p = 0.48 > 0.05)
                        No change in performance detected.
Found 11 outliers among 100 measurements (11.00%)
  11 (11.00%) high severe

bench_retain_mut_none_small
                        time:   [12.992 ns 13.038 ns 13.088 ns]
                        change: [−0.0966% +0.7096% +1.4743%] (p = 0.08 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  5 (5.00%) high mild
  5 (5.00%) high severe
```

Thanks you for maintaining this library.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,6 +44,9 @@ jobs:
       - name: Cargo test w/ malloc_size_of
         run: cargo test --verbose --features malloc_size_of
 
+      - name: Cargo fmt
+        run: cargo --all -- --check
+
       - name: Cargo check w/o default features
         if: matrix.toolchain == 'nightly'
         run: cargo check --verbose --no-default-features
@@ -92,4 +95,3 @@ jobs:
       - name: Mark the job as unsuccessful
         run: exit 1
         if: "!success()"
-
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,9 +44,6 @@ jobs:
       - name: Cargo test w/ malloc_size_of
         run: cargo test --verbose --features malloc_size_of
 
-      - name: Cargo fmt
-        run: cargo --all -- --check
-
       - name: Cargo check w/o default features
         if: matrix.toolchain == 'nightly'
         run: cargo check --verbose --no-default-features
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1226,15 +1226,16 @@ impl<T, const N: usize> SmallVec<T, N> {
 
     #[inline]
     pub fn pop(&mut self) -> Option<T> {
-        if self.is_empty() {
+        let len = self.len();
+        if len == 0 {
             None
         } else {
-            let len = self.len() - 1;
-            // SAFETY: len < old_len since this can't overflow, because the old length is non zero
-            unsafe { self.set_len(len) };
+            let new_len = len - 1;
+            // SAFETY: new_len < len since len is non-zero
+            unsafe { self.set_len(new_len) };
             // SAFETY: this element was initialized and we just gave up ownership of it, so we can
             // give it away
-            let value = unsafe { self.as_mut_ptr().add(len).read() };
+            let value = unsafe { self.as_mut_ptr().add(new_len).read() };
             Some(value)
         }
     }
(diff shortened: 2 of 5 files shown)


fereidani:
Forgot to add that rework of `reserve` is based on std implementation of `reserve`.

.github/workflows/main.yml
```diff
@@ -44,6 +44,9 @@ jobs:
       - name: Cargo test w/ malloc_size_of
         run: cargo test --verbose --features malloc_size_of
 
+      - name: Cargo fmt
+        run: cargo --all -- --check
```
TDecking, line 48:
```suggestion
        run: cargo fmt --all -- --check
```

src/lib.rs
```diff
@@ -1311,16 +1328,26 @@ impl<T, const N: usize> SmallVec<T, N> {
     }
 
     #[inline]
+    #[track_caller]
```
TDecking, line 1331:
```suggestion
```

Rust actually [removed](https://github.com/rust-lang/rust/pull/147042) `#[track_caller]` from most
allocating methods, so it might be better not to add them.

src/lib.rs
```diff
@@ -1332,22 +1359,30 @@ impl<T, const N: usize> SmallVec<T, N> {
...
-            self.try_grow(new_capacity)
-        } else {
-            Ok(())
+            self.try_grow(new_capacity)?;
         }
+        Ok(())
     }
 
     #[inline]
+    #[track_caller]
```
TDecking, line 1368:
```suggestion
```

Same as above.

.github/workflows/main.yml
fereidani: I'm stupid sorry😂!

src/lib.rs
alejandro-vaz: yeah they should be removed in the future

we can keep it for now and we'll remove them on another PR

alejandro-vaz (approved):
& tests pass, miri doesn't complain

.github/workflows/main.yml
alejandro-vaz: this has to be fixed before merge

alejandro-vaz (changes requested):
CI will fail, we can't push this

alejandro-vaz (approved):

alejandro-vaz:
yeah having `cargo fmt` there wouldn't have made much sense anyway, it's not modifying the source

fereidani:
I usually put it there to avoid merging before cargo fmt in my own libraries. but I rushed it in here and code of the workflow was broken. Please consider adding that to the workflow so you can keep the crate formatted based on your required standard.

alejandro-vaz:
note: CI runs are pending

emilio (approved):

PR #397, opened 2025-12-15
Exclude development script from cargo packaging

weiznich:
During a dependency review we noticed that the smallvec crate includes a development script. This development script shouldn't be there as they might, at some point become problematic. As of now they prevent any downstream user from enabling the `[bans.build.interpreted]` option of cargo deny.

I opted for using an explicit include list instead of an exclude list to prevent these files from beeing included in the published packages to make sure that everything that's included is an conscious choice.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ keywords = ["small", "vec", "vector", "stack", "no_std"]
 categories = ["data-structures"]
 readme = "README.md"
 documentation = "https://docs.rs/smallvec/"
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
I'll give others a chance to review before merging, but LGTM.

We'll probably also want to do this for the `v2` branch.

mrobinson (approved):
Makes sense to me.

weiznich:
@mrobinson @nicoburns Is there anything needed from my side to get this PR merged?

mrobinson:
@weiznich Do you mind also making this pull request against the v2 branch?

PR #402, opened 2026-02-12
Remove unnecessary unsafe.

jdm:


Taym95 (approved):

jdm:
Sigh. Not unsafe in newer rust versions, unsafe in our MSRV.

alejandro-vaz:
yeah that unsafe is going to be eternal

managed to remove the macro though in #421 by using rust raw pointer syntax: `unsafe {&raw const/mut self.inline}`

which needs to be unsafe anyway because of union access (weirdly, not in 1.100, but it does in 1.83)

this pr should be closed

PR #406, opened 2026-03-29
Added SmallVec::push_mut andSmallVec::insert_mut

mematthias:
The `push_mut` feature (https://github.com/rust-lang/rust/issues/135974) has already been merged and is expected to be included in the Rust 1.95 stable release. For this reason, we should also add `push_mut` and `insert_mut` here.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1505,24 +1505,33 @@ impl<T, const N: usize> SmallVec<T, N> {
 
     #[inline]
     pub fn insert(&mut self, index: usize, value: T) {
+        _ = self.insert_mut(index, value);
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn insert_mut(&mut self, index: usize, value: T) -> &mut T {
         let len = self.len();
         assert!(
             index <= len,
             "insertion index (is {index}) should be <= len (is {len})"
         );
         self.reserve(1);
-        let ptr = self.as_mut_ptr();
+        let mut ptr = self.as_mut_ptr();
         unsafe {
             // the elements at `index + 1..len + 1` are now initialized
             if index < len {
                 copy(ptr.add(index), ptr.add(index + 1), len - index);
             }
             // the element at `index` is now initialized
-            ptr.add(index).write(value);
+            ptr = ptr.add(index);
+            ptr.write(value);
 
             // SAFETY: all the elements are initialized
             self.set_len(len + 1);
         }
+        let result = unsafe { ptr.as_mut() };
+        unsafe { result.unwrap_unchecked() }
     }
 
     #[inline]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1214,13 +1214,15 @@ impl<T, const N: usize> SmallVec<T, N> {
             self.reserve(1);
         }
 
-        // SAFETY: both the input and output are within the allocation
+        // SAFETY: `len < capacity` after the reserve,
+        //         so the offset stays in bounds of the allocation.
         let ptr = unsafe { self.as_mut_ptr().add(len) };
-        // SAFETY: we allocated enough space in case it wasn't enough,
-        //         so the address is valid for writes.
+
+        // SAFETY: the slot at `len` is free and properly aligned for `T`.
         unsafe { ptr.write(value) };
 
-        // unsafe { self.set_len(len + 1) };
+        // SAFETY: all elements in `0..len + 1` are initialized.
+        //unsafe { self.set_len(len + 1) };
         {
             // This block is an exact copy of `self.set_len`.
             // We have to do this so that Miri doesn't report a "Stacked Borrows" rule violation.
@@ -1232,6 +1234,8 @@ impl<T, const N: usize> SmallVec<T, N> {
             self.len = TaggedLen::new(new_len, on_heap);
         }
 
+        // SAFETY: `ptr` is aligned, non-null and points to the element initialized above;
+        //         the borrow is tied to `&mut self`, so it is exclusive.
         unsafe { &mut *ptr }
     }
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1218,21 +1220,34 @@ impl<T, const N: usize> SmallVec<T, N> {
...
         }
+
         // SAFETY: both the input and output are within the allocation
         let ptr = unsafe { self.as_mut_ptr().add(len) };
         // SAFETY: we allocated enough space in case it wasn't enough, so the address is valid for
         // writes.
         unsafe { ptr.write(value) };
         unsafe { self.set_len(len + 1) }
+
+        let result = unsafe { ptr.as_mut() };
```
emilio: Maybe just `&mut *ptr`? Or why would `as_mut().unwrap_unchecked()` be preferred?

emilio (commented):
Would be nice to split the formatting into a separate (preliminary?) PR.

src/lib.rs
mematthias: That is just my writing style; it makes it easier to search.
I simplified it anyway

emilio (approved):

emilio (dismissed):
Miri seems to be complaining.

balt-dev:
miri seems to be getting confused on the provenance?

balt-dev:
as the guy who did push_mut in std, i feel like the change just made kind of defeats the point?  the whole point of push_mut is that you're _not_ rederiving the pointer
i say the old version is fine, we just need some way to tell miri it's okay

mematthias:
@balt-dev While I understand the concerns raised, I am currently unable to resolve the Miri report. That said, this should not block us from shipping the feature. At the moment, users neither benefit from nor are harmed by it. But by making it available now, we ensure that once the optimization is in place, existing users can immediately take advantage of it without any disruption to their experience.

One possible (but dirty) solution would be to use `set_len` directly (inline) within the `push_mut` function:
```rust
#[inline]
pub fn push(&mut self, value: T) {
    _ = self.push_mut(value);
}

#[inline]
#[must_use]
pub fn push_mut(&mut self, value: T) -> &mut T {
    let len = self.len();
    if len == self.capacity() {
        self.reserve(1);
    }
    // SAFETY: both the input and output are within the allocation
    let ptr = unsafe { self.as_mut_ptr().add(len) };
    // SAFETY: we allocated enough space in case it wasn't enough, so the address is valid for
    // writes.
    unsafe { ptr.write(value) };

    // unsafe { self.set_len(len + 1) };
    {
        let new_len = len + 1;
        debug_assert!(new_len <= self.capacity());
        let on_heap = self.len.on_heap();
        self.len = TaggedLen::new(new_len, on_heap);
    }

    unsafe { &mut *ptr }
}
```

alejandro-vaz:
> @mematthias @balt-dev While I understand the concerns raised, I am currently unable to resolve the Miri report. That said, this should not block us from shipping the feature. At the moment, users neither benefit from nor are harmed by it. But by making it available now, we ensure that once the optimization is in place, existing users can immediately take advantage of it without any disruption to their experience.
> 
> One possible (but dirty) solution would be to use `set_len` directly (inline) within the `push_mut` function:
> 
> ```rust
> #[inline]
> pub fn push(&mut self, value: T) {
>     _ = self.push_mut(value);
> }
> 
> #[inline]
> #[must_use]
> pub fn push_mut(&mut self, value: T) -> &mut T {
>     let len = self.len();
>     if len == self.capacity() {
>         self.reserve(1);
>     }
>     // SAFETY: both the input and output are within the allocation
>     let ptr = unsafe { self.as_mut_ptr().add(len) };
>     // SAFETY: we allocated enough space in case it wasn't enough, so the address is valid for
>     // writes.
>     unsafe { ptr.write(value) };
> 
>     // unsafe { self.set_len(len + 1) };
>     {
>         let new_len = len + 1;
>         debug_assert!(new_len <= self.capacity());
>         let on_heap = self.len.on_heap();
>         self.len = TaggedLen::new(new_len, on_heap);
>     }
> 
>     unsafe { &mut *ptr }
> }
> ```

yeah

when `set_len` is called it takes a mutable reference into the full self and then miri can't prove that the reference is still pointing to valid data because the location it points to might have been mutated since and thus it might be uninitialized data (it might have been freed => returns null pointer to user => use after free). even though that isn't what happens in our code in any case

the `TaggedLen` / `set_len` model is not a very good abstraction for what we are trying to do

I honestly support setting the length inline with `self.len = /*...*/`

the `TaggedLen` thing should probably be redesigned anyway to be honest

but before hacking it we should change how length behaves, then we implement it properly

mematthias:
@alejandro-vaz Should the same dirty workaround be used for `insert_mut`?

alejandro-vaz:
yeah

the `set_len` method doesn't make much sense and just confuses the borrow checker by taking a mutable reference of the whole smallvec instead of just its len field

alejandro-vaz:
please rebase the commit history onto servo:v2 to remove merge commits and have a linear history

src/lib.rs
```diff
@@ -1200,24 +1198,45 @@ impl<T, const N: usize> SmallVec<T, N> {
...
         let ptr = unsafe { self.as_mut_ptr().add(len) };
-        // SAFETY: we allocated enough space in case it wasn't enough, so the address is valid for
-        // writes.
+
+        // SAFETY: the slot at `len` is free and properly aligned for `T`.
         unsafe { ptr.write(value) };
-        unsafe { self.set_len(len + 1) }
+
+        // SAFETY: all elements in `0..len + 1` are initialized.
+        //unsafe { self.set_len(len + 1) };
```
alejandro-vaz: looks like this should be removed

src/lib.rs
```diff
@@ -1200,24 +1198,45 @@ impl<T, const N: usize> SmallVec<T, N> {
...
+        {
+            // This block is an exact copy of `self.set_len`.
+            // We have to do this so that Miri doesn't report a "Stacked Borrows" rule violation.
+            // See PR/406
+
+            let new_len = len + 1;
+            debug_assert!(new_len <= self.capacity());
+            let on_heap = self.len.on_heap();
+            self.len = TaggedLen::new(new_len, on_heap);
+        }
```
alejandro-vaz, line 1271: yeah this is good for now until we implement #451

alejandro-vaz (changes requested):

src/lib.rs
mematthias: I left that in on purpose, but I can remove it.

alejandro-vaz: I think we should remove it, because that method would disappear anyway with #451 

the method itself existing is the issue, because it takes a mutable reference into the smallvec object as a whole instead of only the len field

alejandro-vaz (approved):

alejandro-vaz:
@mematthias I had to merge upstream changes, it annoyed rustfmt

please run `cargo fmt --all` so I can approve it

PR #407, opened 2026-04-08
Fix use-after-free in DrainFilter::keep_rest for SmallVec with capacity 0

Shnatsel:
In current v1, `keep_rest()` checks `mem::size_of::<T>()` (the Array type) instead of `mem::size_of::<T::Item>()` (the element type) to decide whether to backshift tail elements. For `SmallVec<[T; 0]>`, the array size is always 0, so the backshift is skipped and subsequent accesses read freed memory.

This PR adds a test that triggers the issue under miri, reproducible by `cargo +nightly miri test`, and then applies the one-line fix in a subsequent commit.

This does **not** appear to be exploitable in practice since it requires activating a non-default Cargo feature combined with SmallVec with inline capacity 0. Using capacity 0 is rather silly - you should just use `std::Vec` at that point - so I don't expect this configuration to be common in the wild.

The issue was autonomously discovered by Claude Opus 4.6, but I have manually verified the PoC and the fix.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -602,7 +602,7 @@ where
 
         unsafe {
             // ZSTs have no identity, so we don't need to move them around.
-            let needs_move = mem::size_of::<T>() != 0;
+            let needs_move = mem::size_of::<T::Item>() != 0;
 
             if needs_move && this.idx < this.old_len && this.del > 0 {
                 let ptr = this.vec.as_mut_ptr();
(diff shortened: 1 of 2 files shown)


jdm (approved):
Thank you!

Shnatsel:
The CI failure is unrelated to the changes in this PR, and is also present on the tip of the v1 branch.

Shnatsel:
I see this PR is approved, is there anything blocking merging it? Even if this is not urgent, I'd love to see the fix shipped sooner rather than later.

emilio (approved):

PR #410, opened 2026-06-11
Work around rustc 1.93 perf regression with MaybeUninit

### LINKED ISSUE #408: Performance regression with rustc 1.93+
Starting with rustc 1.93, wrapping a SmallVec in any struct causes rustc to emit code that zeroes the entire struct, instead of only touching the capacity field (which was the behavior until rustc 1.92).
The larger the Array in the SmallVec, the bigger the impact. In the real world, this caused a noticeable regression in a perf test in Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=2046376)
The rustc regression itself is filed here: https://github.com/rust-lang/rust/issues/157743

This can be trivially reproduced by adding the following benchmark to benches/bench.rs:
```rust
#[bench]
fn bench_new_in_newtype(b: &mut Bencher) {
    struct Wrap(SmallVec<[u64; 32]>);
    b.iter(|| test::black_box(Wrap(SmallVec::new())))
}
```

On my machine the benchmark results look like this:
| | nightly-2025-11-23 (last good) | nightly-2025-11-24 (first bad) |
|-|----------------------------|------------------|
|union| 7.63ns | 10.74ns |
|non-union| 8.70ns | 13.04ns |

That is a significant overhead, and in a tight loop as in the case of Firefox, can lead to a large regression.

glandium:
Fixes #408

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -651,6 +651,19 @@ impl<A: Array> SmallVecData<A> {
             inline: core::mem::ManuallyDrop::new(inline),
         }
     }
+    // Workaround for https://github.com/rust-lang/rust/issues/157743: when from_inline is
+    // called with MaybeUninit::uninit(), rustc 1.93+ GVN propagates const <uninit> into the
+    // ManuallyDrop::new() aggregate, causing LLVM to materialize a global constant that
+    // MemCpyOpt then collapses into a memset over the whole struct. Using assume_init() of a
+    // doubly-wrapped MaybeUninit produces Immediate::Uninit instead of const <uninit>, which
+    // codegen handles as undef without emitting any global. This function also avoids
+    // introducing an intermediate local that would inflate stack frames in debug builds.
+    #[inline]
+    fn empty() -> SmallVecData<A> {
+        // SAFETY: ManuallyDrop<MaybeUninit<A>> is valid for any bit pattern including
+        // uninitialized bytes, so assume_init() on a MaybeUninit of that type is sound.
+        SmallVecData { inline: unsafe { MaybeUninit::uninit().assume_init() } }
+    }
     #[inline]
     unsafe fn into_inline(self) -> MaybeUninit<A> {
         core::mem::ManuallyDrop::into_inner(self.inline)
@@ -712,6 +725,13 @@ impl<A: Array> SmallVecData<A> {
     #[inline]
     fn from_inline(inline: MaybeUninit<A>) -> SmallVecData<A> {
         SmallVecData::Inline(inline)
+    }
+    // See the comment on the union variant's empty() for why this exists.
+    #[inline]
+    fn empty() -> SmallVecData<A> {
+        // SAFETY: MaybeUninit<A> is valid for any bit pattern including uninitialized bytes,
+        // so assume_init() on a MaybeUninit of that type is sound.
+        SmallVecData::Inline(unsafe { MaybeUninit::uninit().assume_init() })
     }
     #[inline]
     unsafe fn into_inline(self) -> MaybeUninit<A> {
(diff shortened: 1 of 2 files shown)


emilio (commented):
Some of the unsafe here seems gratuitous. Can you apply:

```diff
diff --git a/src/lib.rs b/src/lib.rs
index f7a2dff..446d736 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -653,7 +653,7 @@ impl<A: Array> SmallVecData<A> {
         // memset over the whole struct. Initializing the field via a raw pointer write breaks
         // the GVN constant-propagation chain and restores the expected codegen.
         let mut data = SmallVecData {
-            inline: unsafe { MaybeUninit::uninit().assume_init() },
+            inline: core::mem::ManuallyDrop::new(MaybeUninit::uninit()),
         };
         unsafe { core::ptr::addr_of_mut!(*data.inline).write(inline) };
         data
@@ -723,7 +723,7 @@ impl<A: Array> SmallVecData<A> {
         // `MaybeUninit::uninit()` as a global constant, which LLVM then collapses into a
         // memset over the whole struct. Constructing with a fresh uninit payload first and
         // then overwriting via a raw pointer write breaks the GVN chain.
-        let mut data = SmallVecData::Inline(unsafe { MaybeUninit::uninit().assume_init() });
+        let mut data = SmallVecData::Inline(MaybeUninit::uninit());
         match &mut data {
             SmallVecData::Inline(a) => unsafe { core::ptr::write(a as *mut _, inline) },
             _ => unsafe { core::hint::unreachable_unchecked() },
```

Looks good to me with that.

glandium:
I'm not sure why this is not happening with cargo bench, but the more broad test reduced from Firefox still does trigger the bug with your suggestion: https://rust.godbolt.org/z/Yq4rasTx4

src/lib.rs
```diff
@@ -647,9 +647,16 @@ impl<A: Array> SmallVecData<A> {
...
-        SmallVecData {
-            inline: core::mem::ManuallyDrop::new(inline),
-        }
+        // Workaround for https://github.com/rust-lang/rust/issues/157743: constructing the
+        // union directly with `ManuallyDrop::new(inline)` causes rustc 1.93+ to materialize
+        // `MaybeUninit::uninit()` as a global constant, which LLVM then collapses into a
+        // memset over the whole struct. Initializing the field via a raw pointer write breaks
+        // the GVN constant-propagation chain and restores the expected codegen.
+        let mut data = SmallVecData {
+            inline: unsafe { MaybeUninit::uninit().assume_init() },
```
nicoburns: Why is it valid to call `assume_init()` here? Docs (https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#safety) say that calling `assume_init()` on uninitialized data is immediate UB. And to me it looks like this data could be uninitialized?

glandium: we're overwriting the data immediately.

nicoburns: I might be wrong, but I thought you had to overwrite the data *before* calling `assume_init()` (and that this is the purpose of `MaybeUninit`: to allow you to have write access to an uninitialized buffer (so that you can initialize it) without that being UB)

emilio: @nicoburns This also tripped me up when I looked at it, but it seems fine. It's not about overwriting the data, IMO. The types here are: `MaybeUninit<ManuallyDrop<MaybeUninit<A>>>::uninit().assume_init()`.

It's always safe to `assume_init` a `ManuallyDrop<MaybeUninit<..>>`... It should be redundant, but see the discussion above, it seems to matter.

emilio (approved):
Pretty gross, but I guess it's fine. Once this is fixed in a stable Rustc we should revert the change of course...

src/lib.rs
nicoburns: Ah, there's an inner `MaybeUninit`? Yeah, I think that makes it fine. I would love to see a SAFETY comment to that effect if possible.

emilio (approved):
r=me

PR #411, opened 2026-06-11
Fix Windows CI.

emilio:
This example doesn't need such a big inline space.

### DIFF
--- a/tests/debugger_visualizer.rs
+++ b/tests/debugger_visualizer.rs
@@ -4,6 +4,7 @@ use smallvec::{smallvec, SmallVec};
 #[inline(never)]
 fn __break() {}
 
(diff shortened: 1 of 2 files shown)


tests/debugger_visualizer.rs
```diff
@@ -4,6 +4,7 @@ use smallvec::{smallvec, SmallVec};
 #[inline(never)]
 fn __break() {}
 
+/* Disabled, see https://github.com/servo/rust-smallvec/issues/413
```
emilio, line 7: I tried `#[ignore]` but that didn't work.

emilio:
I'm going to go ahead and merge this because Windows CI has been failing for quite a while.

Issue #413, opened 2026-06-11
Debugger visualizer test is failing.


emilio:
https://github.com/servo/rust-smallvec/actions/runs/27339093433/job/80770932750?pr=411

I think we should disable it for now.

emilio:
This was added in #286, cc @ridwanabdillahi

It fails on trunk and it prevents patches from landing, so disabling for now.

Issue #415, opened 2026-06-11
smallvec 1.15.2 doc build failed


atouchet:
https://docs.rs/crate/smallvec/latest/builds/3549516

```
[INFO] [stderr] warning: target filter specified, but no targets matched; this is a no-op
[INFO] [stderr]  Documenting smallvec v1.15.2 (/opt/rustwide/workdir)
[INFO] [stderr] error: couldn't read src/../debug_metadata/smallvec.natvis: No such file or directory (os error 2)
[INFO] [stderr]    --> src/lib.rs:100:25
[INFO] [stderr]     |
[INFO] [stderr] 100 |     debugger_visualizer(natvis_file = "../debug_metadata/smallvec.natvis")
[INFO] [stderr]     |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[INFO] [stderr] 
[INFO] [stderr] error: could not document `smallvec`
```

TimNN:
It's not just the doc build, it's any build enabling the `debugger_visualizer` feature, presumably since #397.

alejandro-vaz:
solved with #420, this issue should be closed

atouchet:
A new version of smallvec should still be published.

alejandro-vaz:
@atouchet agree

PR #420, opened 2026-08-20
feat: removed debugger visualizer + tests

alejandro-vaz:
Removed the `debugger_visualizer` feature and test causing #413 and #415.

The test case was previously commented out to pass tests.

Feature is no longer being maintained.

Docs generate successfully with `cargo doc --all-features`.

Removed feature from `Cargo.toml`.

### Before merging

Figure out whether we really want to remove the feature or only the testing for it (i.e. assume the `.natvis` script written long ago is correct).


emilio (approved):
I think given this feature broke, nobody noticed except for #415, and I can't test it, it seems fine to remove.

We could keep the natvis file I guess, but I have very low confidence it has stayed correct with the change to `union` and so on.

alejandro-vaz:
apparently there was support for a workflow test that ran the debugger visualizer case with the union feature

which probably also failed but never got to run on the logs because the one without the union feature failed first

alejandro-vaz:
PR is ready to be merged

alejandro-vaz:
closes #413 and #415

Issue #419, opened 2026-08-20
Remove `Cargo.lock` file exception from `.gitignore`


alejandro-vaz:
As recommended by https://github.com/rust-lang/cargo/pull/12382.

Lockfiles should be committed even in libraries.

Affects `v1` and `v2` branch.

alejandro-vaz:
merged in #420

alejandro-vaz:
v2 in #421

alejandro-vaz:
v1 won't implement this due to `Cargo.lock` version incompatibility

v2 is safe, lockfile v4 was made the default in 1.83

can't be implemented into v1 because it breaks the CI workflow

alejandro-vaz:
merged

PR #421, opened 2026-08-20
chore: small v2 cleanup

alejandro-vaz:
- added `Cargo.lock` to filetree (see #419)
- removed address-getting macro
- replaced deprecated const `std::usize::MAX`

## removed address-getting macro

replaced `addr_of!(/*...*/)` pointer getting inline functions to idiomatic `&raw const /*...*/` syntax introduced in Rust 1.82.

This is compatible with our MSRV (1.83)

## deprecated const

replaced the deprecated const `std::usize::MAX` in `tests.rs` to the associated const `usize::MAX`

it was stabilized in rust 1.43


alejandro-vaz:
ready for review

emilio (approved):

PR #423, opened 2026-08-20
fix: removed Cargo.lock from .gitignore

alejandro-vaz:
see #419 

note: there are duplicated commits already merged but it shouldn't be an issue


alejandro-vaz:
this is never going to work

1.36 uses different Carg.lock file version and converting between them is a nightmare

we should let it be like this

Issue #418, opened 2026-07-26
Reduce bloat from `smallvec!` in the inline case


bolshoytoster:
Currently, `smallvec!`'s inline case is handled [by repeatedly pushing to a `SmallVec`](https://github.com/servo/rust-smallvec/blob/bc8a854926a8d940164f6c4ad4fc6efe51962e93/src/lib.rs#L2678). I've measured that this produces a lot of LLVM IR + bloat in the output binary.

I've managed to reduce the size in the inline case by using `From<[T, M]>`:
```rust
($($x:expr),+$(,)?) => ({
    $crate::SmallVec::from([$($x,)+])
});
```
(Note: for that to have an effect, `From<[T, M]>` needs to be marked `#[inline]` as well)

I haven't measured how this affects the heap case, but if it's a concern, I've found that something like this reduces the output to the same size, but is a little more verbose:
```rust
($($x:expr),+$(,)?) => ({
    const COUNT: usize = 0usize $(+ $crate::smallvec!(@one $x))+;
    let vec = $crate::SmallVec::new();
    if false {
        vec
    } else if COUNT <= vec.capacity() {
        core::mem::forget(vec);
        $crate::SmallVec::from_buf([$($x,)+])
    } else {
        core::mem::forget(vec);
        $crate::SmallVec::from_vec($crate::alloc::vec![$($x,)+])
    }
});
```

alejandro-vaz:
what is the reasoning behind `if false`??

and forgetting vec in the two possible branches??

I'm planning to rework the macro, but I don't understand what's going on

bolshoytoster:
Sorry, I probably should've explained:
The
```rs
if false {
    vec
}
```
is needed so the compiler can still infer the type of `vec` (it's only used to get the inferred capacity). There's probably a better way to do that.

Forgetting the vec is needed because otherwise the compiler inserts drop code for it, even though its empty, so this reduces bloat.

alejandro-vaz:
yeah but that doesn't make much sense

forall `COUNT` there exists an `N` such that `N == vec.capacity()` and `COUNT <= N`

which makes the type `SmallVec<[T; N]>`

think about it this way: we are always passing an array with a finite amount of elements to `smallvec!`, thus there always exists some `N` such that our array is of type `[T; N]`

[...]

PR #424, opened 2026-08-20
Reduce bloat from `smallvec!` in the inline case

bolshoytoster:
See #418 for details.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2672,14 +2672,7 @@ macro_rules! smallvec {
         $crate::from_elem($elem, $n)
     });
     ($($x:expr),+$(,)?) => ({
-        const COUNT: usize = 0usize $(+ $crate::smallvec!(@one $x))+;
-        let mut vec = $crate::SmallVec::new();
-        if COUNT <= vec.capacity() {
-            $(vec.push($x);)*
-            vec
-        } else {
-            $crate::SmallVec::from_vec($crate::alloc::vec![$($x,)+])
-        }
+        $crate::SmallVec::from([$($x,)+])
     });
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2667,7 +2667,7 @@ macro_rules! smallvec {
         $crate::from_elem($elem, $n)
     });
     ($($($x:expr),+$(,)?)?) => ({
-        $crate::SmallVec::from([$($($x,)+)?])
+        $crate::SmallVec::from([$($($x),+)?])
     });
 }
 
@@ -2680,7 +2680,7 @@ macro_rules! smallvec_inline {
     });
     ($($x:expr),+ $(,)?) => ({
         const N: usize = 0usize $(+ $crate::smallvec_inline!(@one $x))*;
-        $crate::SmallVec::<_, N>::from_buf([$($x,)*])
+        $crate::SmallVec::<_, N>::from_buf([$($x),*])
     });
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2663,16 +2663,11 @@ impl<T, const N: usize> core::iter::FromIterator<T> for SmallVec<T, N> {
 
 #[macro_export]
 macro_rules! smallvec {
-    // count helper: transform any expression into 1
-    (@one $x:expr) => (1usize);
-    () => (
-        $crate::SmallVec::new()
-    );
     ($elem:expr; $n:expr) => ({
         $crate::from_elem($elem, $n)
     });
-    ($($x:expr),+$(,)?) => ({
-        $crate::SmallVec::from([$($x,)+])
+    ($($($x:expr),+$(,)?)?) => ({
+        $crate::SmallVec::from([$($($x,)+)?])
     });
 }
 


src/lib.rs
```diff
@@ -2672,14 +2672,7 @@ macro_rules! smallvec {
         $crate::from_elem($elem, $n)
     });
     ($($x:expr),+$(,)?) => ({
```
alejandro-vaz: shouldn't this be instead

```rust
($($x:expr),* $(,)?) => // ...
```

src/lib.rs
alejandro-vaz: and then the `() => // ...` branch can disappear because the glob pattern of the one with elements it would cover nullable tokenstreams

alejandro-vaz: this branch can be removed if completely

alejandro-vaz, line 2675: something similar should be done here, removing `@one` recursion

alejandro-vaz (changes requested):

src/lib.rs
bolshoytoster: Thanks, I missed that.

src/lib.rs
alejandro-vaz: though it wouldn't make sense calling `smallvec[,]`, and it actually isn't a valid array, so it should probably be

```rust
($($($x:expr),+ $(,)?)?) => // ...
```

src/lib.rs
bolshoytoster: Would that change the behavior of the macro? It currently outputs a `SmallVec` with the exact capacity for the items passed. If I removed the `const N: usize = 0usize $(+ $crate::smallvec_inline!(@one $x))*;` and just had it infer `N`, then it would be identical to the other macro.

Maybe something like
```rs
    ($($x:expr),+ $(,)?) => ({
        let buf = [$($x,)*];
        $crate::SmallVec::from_buf_and_len(buf, buf.len())
    });
```
would work?

alejandro-vaz: actually to me it how it is written now seems wrong

`smallvec_inline!` should basically be the parity of `smallvec!` but all being inline

so I honestly think it should be 

```rust
    ($($x:expr),+ $(,)?) => ({
        $crate::SmallVec::from_buf([$($x),*])
        // from_buf::<S>([_; N]) lets S be inferred from the caller
        // so:
        // let s: SmallVec<usize, 4> = smallvec_inline![1, 2]
    });
```

alejandro-vaz: currently it is counting so it doesn't allow to get from context the desired maximum capacity, but that's not what `smallvec!` does

and they should probably behave equivalently

src/lib.rs
```diff
@@ -2663,16 +2663,11 @@ impl<T, const N: usize> core::iter::FromIterator<T> for SmallVec<T, N> {
...
-    () => (
-        $crate::SmallVec::new()
-    );
     ($elem:expr; $n:expr) => ({
         $crate::from_elem($elem, $n)
     });
-    ($($x:expr),+$(,)?) => ({
-        $crate::SmallVec::from([$($x,)+])
+    ($($($x:expr),+$(,)?)?) => ({
+        $crate::SmallVec::from([$($($x,)+)?])
```
alejandro-vaz: `$crate::SmallVec::from([$($($x,)+)?])`

should be

`$crate::SmallVec::from([$($($x),+)?])`

weird rust quirk in which the outside comma means there is no trailing comma

probably doesn't matter much but it matches the branch signature

src/lib.rs
bolshoytoster: Would that be a breaking change? And if we do that, we may as well just remove `smallvec_inline!` in favor of `smallvec!`.

Actually, looking at the blame, it looks like `smallvec_inline!` came from #265 (2021) and might just be obselete now.

It would still be a breaking change to remove it, so maybe it could just forward to `smallvec!`?

src/lib.rs
bolshoytoster: I left the comma inside for consistency because [the original did it in the heap path](https://github.com/servo/rust-smallvec/blob/v2/src/lib.rs#L2681). `smallvec_inline` also has the comma inside. I can change it if you prefer it outside.

src/lib.rs
alejandro-vaz: smallvec v2 is on alpha so breaking changes aren't an specially big deal if we get the architecture right

I mean, yeah, the breaking change would be that anyone who is using `smallvec_inline!` couldn't be relying on the macro for inferring the value of `N`

it doesn't seem intentional to me that they let constant fixed

but it is true that we could just remove the macro entirely then

but what use case does `smallvec_inline!` have?? basically none. just use `smallvec!` and annotate or (most of the time) the compiler will infer the value you want

I'd lean now towards removing the `smallvec_inline!` macro altogether

I'll request a second reviewer later, but I think this is the right direction for such a small macro whose use case is negligible and was meant for a hack 5 years ago

src/lib.rs
alejandro-vaz: yeah remove it everywhere you see it

why would anyone want a trailing comma in macro_rules generation everywhere

hahahaha it's so tiny it's laughable but it doesn't make sense

like if it had been intentionally made that way

src/lib.rs
alejandro-vaz: remove it (and if it has any uses in the codebase simply replace) and the pr is ready

bolshoytoster: There may be a use, actually. `smallvec_inline` can be used in const, but `From<[T, M]>` (and therefore `smallvec!`) isn't const.

alejandro-vaz: you know what, I think we should leave `smallvec_inline!` existing for now but with the `::from_buf` of above

because the more you think about it, the more you realize that the `smallvec!` and `smallvec_inline!` macros are just unnecessary macros around `SmallVec::from` and `SmallVec::from_buf`

the rust guidelines are very simple: don't make a macro out of which can be done as a function, yet this is what these macros seem to be doing

I'll annotate this, maybe I'll make an issue in a few days with the worklist for v2, but I'm very against having a macro with seemingly no good use case other than syntactic sugar

we'll also have to check what #396 did regarding this (if anything) as well

alejandro-vaz: > you know what, I think we should leave `smallvec_inline!` existing for now but with the `::from_buf` of above
> 
> because the more you think about it, the more you realize that the `smallvec!` and `smallvec_inline!` macros are just unnecessary macros around `SmallVec::from` and `SmallVec::from_buf`
> 
> the rust guidelines are very simple: don't make a macro out of which can be done as a function (literally what the book says). yet this is what these macros seem to be doing
> 
> I'll annotate this, maybe I'll make an issue in a few days with the worklist for v2, but I'm very against having a macro with seemingly no good use case other than syntactic sugar
> 
> we'll also have to check what #396 did regarding this (if anything) as well

this is what's left

#396 did not touch those macros

alejandro-vaz: this discussion will be continued on #432

alejandro-vaz (approved):

emilio (approved):
Should v1 get this as well?

bolshoytoster:
> Should v1 get this as well?

I'm not sure if this will beneft v1, although it probably will

Issue #425, opened 2026-08-21
Tracking issue for v2


alejandro-vaz:
> Information about the state of v2 will be here until a proper project/milestone is created
>
> for the v2 spec, see [the wiki](https://github.com/servo/rust-smallvec/wiki)

## Relevant PRs

- #431: awaiting author changes
- #396: not yet reopened after being reverted by #430 due to MSRV conflicts
- #391: postponed
- #387: proof of concept, awaiting author changes

## `2.0.0-alpha.13`

Remaining:
- #427 
- [ ] `2.0.0-alpha.12`..`2.0.0-alpha.13` diff review

alejandro-vaz:
let's migrate to proper milestones

Issue #398, opened 2025-12-15
Drop bincode support entirely for smallvec 2.0


valadaptive:
The current maintainer of bincode [has explicitly stated that the project is not maintained anymore](https://old.reddit.com/r/rust/comments/1pmw2c0/whats_going_on_with_bincode/nu4gc98/), and nobody should be filing issues on it:

> If you're so bothered by this just use the 1.x bincodes. Seriously. It's done. For years any issues always turned out to be user error. There's never once been a CVE and if there is we trust that the person finding it will be smart enough to find us. (And also less of a tax on our mental health).
> 
> Filing issues for 2.x? Don't. I don't care. It's bincode. It's dead simple. I'll get around to finding the issues myself. Maybe. When I get back into programming. Eventually.

As such, I think that bincode support should *not* be forward-ported to smallvec 2, as we should be encouraging the ecosystem to migrate away from it and use other serialization frameworks.

alejandro-vaz:
merged into v2

PR #426, opened 2026-08-21
feat: drop bincode entirely

alejandro-vaz:
- dropped support for `bincode` entirely (#398)
- added readme notice about #425 
- refactored `Cargo.toml` `include => exclude` to make exclusions explicit

## bincode drop

replaced test with `serde_json` instead

for the future: #428

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,25 +3,28 @@
 version = 4
 
 [[package]]
-name = "bincode"
-version = "1.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
-dependencies = [
- "serde",
-]
-
-[[package]]
 name = "bytes"
 version = "1.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


alejandro-vaz:
ready for review

README.md
```diff
@@ -13,6 +14,7 @@ rust-smallvec
 [#183]: https://github.com/servo/rust-smallvec/issues/183
 [#240]: https://github.com/servo/rust-smallvec/issues/240
 [#284]: https://github.com/servo/rust-smallvec/issues/284
+[#284]: https://github.com/servo/rust-smallvec/issues/425
```
emilio: `[#425]` right?

Cargo.lock
```diff
@@ -2,27 +2,30 @@
 # It is not intended for manual editing.
 version = 4
 
-[[package]]
```
emilio, line 5: This actually removes one dep and adds a lot more? Seems unfortunate?

alejandro-vaz: #428 already plans to remove it, but the priority was to remove bincode because it's unmaintained (#398)

alejandro-vaz:
oh no, this comes from failed CI at #396 

clap requires edition 2024

alejandro-vaz:
this won't succeed until #396 is reverted with #430

emilio (approved):
I guess these are dev-dependencies so it doesn't matter much either way.

PR #430, opened 2026-08-21
Revert "Refactor v2: faster pop/reserve/retain_mut, modern Criterion benches, Miri-friendly tests + fmt in CI"

alejandro-vaz:
Reverts servo/rust-smallvec#396

due to failed CI checks

issue: `clap_lex` requires rust edition 2024, but we are working with 2021

errors when testing 1.83 version

we are reverting to the latest known-good version

## long version

#396 depends on `criterion = "0.8"`, which has a MSRV of 1.86, but the MSRV in our toolchain is 1.83

we will revert #396, bring back the PR and make the necessary changes by downgrading to `criterion = "0.7"` that only needs rust 1.80+

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -92,3 +92,4 @@ jobs:
       - name: Mark the job as unsuccessful
         run: exit 1
         if: "!success()"
+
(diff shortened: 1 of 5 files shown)


alejandro-vaz:
this should be urgently merged

jdm (approved):

Issue #427, opened 2026-08-21
refactor `lib.rs` into modules


alejandro-vaz:
split the `lib.rs` main file into submodules for easier navigation and separation of concerns:

```
lib.rs
rawsmallvec.rs
taggedlen.rs
serde.rs
drain.rs
extractif.rs
...
...
```

the modules should have clear API boundaries, no ad-hoc splitting. e.g.:
- another type with its methods: `TaggedLen`, `RawSmallVec`
- features: `serde`, `specialization`

key rule: a file has one purpose and touches one part of the code

key rule: if you have to make a `mod {/*...*/}` item, then it probably should be its own file

alejandro-vaz:
ideally this should be done after merging #387

PR #431, opened 2026-08-21
Fix warnings

zachs18:
Fixes `unused_unsafe` warnings on Rust >=1.92.0 by adding `#[allow]`s since the relevant code is not safe in the MSRV 1.83.0 declared in Cargo.toml (this could also be fixed by bumping MSRV to 1.92.0 and removing the `unsafe` blocks).

Fixes specialization syntax future-compatibility-warnings by moving the `spec_traits` module to a file (cc #427 ).


alejandro-vaz (approved):

src/spec_traits.rs
alejandro-vaz, line 1: I think the clear API boundary here is not `spec_traits.rs`, I'd honestly say it is:
- `specialization.rs`
- with all the methods that involve specialization
- `lib.rs` with a conditional compilation flag that includes `specialization.rs` only when the feature is enabled
- the normal methods on `SmallVec` on `lib.rs` that also have specialization have a conditional compilation flag of feature disabled

src/lib.rs
```diff
@@ -199,15 +199,17 @@ impl<T, const N: usize> RawSmallVec<T, N> {
     #[inline]
     const fn as_ptr_inline(&self) -> *const T {
         // SAFETY: it is safe because we aren't reading the value, just getting a
-        // reference to it. reading it would be UB potentially, but for that downstream
-        // unsafe is required
-        (unsafe {&raw const self.inline}) as *mut T
+        // pointer to it. reading it would be UB potentially, but for that downstream
+        // unsafe is required. This is no longer unsafe as of Rust 1.92.0.
+        #[allow(unused_unsafe, reason = "requires unsafe in MSRV 1.83.0")]
```
alejandro-vaz, line 204: this is gold

alejandro-vaz (changes requested):

alejandro-vaz:
PR stalled, will implement the warnings thing on modularization

thanks for the idea, that was a great one

PR #438, opened 2026-08-23
refactor: modularize files and refactor existing implementations

alejandro-vaz:
this PR is the preliminary implementation of #427 

## scope of this PR

this PR aims to refactor the codebase to make it navigable 

it starts to split the old 3k LOC `lib.rs` into small submodules based on abstraction boundaries:
- `allocationerror.rs` => error types
- `bytes.rs` => functionality provided by the `bytes` crate, gated by a cfg
- `comparisons.rs` => `Eq` and `Ord`, total and partial
- `lib.rs` => remaining not-yet-refactored code
- `mallocsizeof.rs` => `MallocSizeOf` work gated behind a cfg
- `rawsmallvec.rs` => already implemented by @zachs18 on #348 
- `references.rs` => asref, borrow, deref. for const and mut references
- `serde.rs` => serde's serialize and deserialize gated behind a cfg
- `std.rs` => standard library implementations. currently, only `std::io::Write`
- `taggedlen.rs` => `TaggedLen` and its implementation
- `tests.rs` => tests, already present before

## other changes

- renamed `CollectionAllocErr` to `AllocationError`
- removed an `is_zst<T>()` helper which can be inlined by a simple comparison
- changes `rustfmt.toml` a bit to group imports and removed the use of the `vec!` macro so the formatter didn't get confused about whether it was referencing the `vec` module or the `vec!` macro

> [!NOTE]
> documentation doesn't build properly, see #439

> [!NOTE]
> experimented a bit with `rustfmt.toml`, should make files easier to read now


rustfmt.toml
```diff
@@ -0,0 +1,10 @@
+wrap_comments = true
+imports_granularity = "One"
+group_imports = "One"
+format_code_in_doc_comments = true
+match_arm_blocks = false
+blank_lines_lower_bound = 0
+blank_lines_upper_bound = 0
```
zachs18, line 11: I don't think removing all blank lines improves readability; I think it makes it worse. Also note that `servo/servo` doesn't do this ([example](https://github.com/servo/servo/blob/b4b6d8c6ae73415717fb39e6db610109ea105ff0/ports/servoshell/lib.rs#L92))

alejandro-vaz: fair

then probably 0 for lower bound and 1 for upper I guess. 2+ looks weird to me

did it out of inertia, didn't think much about it, I'll fix it tomorrow

code looks weird though now that you're saying it

alejandro-vaz:
I will create a new branch, this has gotten messy

PR #443, opened 2026-08-24
recovery: #396 from @fereidani

alejandro-vaz:
this PR brings back #396 (ft. @fereidani) after downgrading `criterion` to v0.7 as specified in #430

PR #446, opened 2026-08-25
Fix retain_mut panic and drop behavior

### LINKED ISSUE #444: fix `retain_mut` panic/drop behavior to match vec API
I reviewed the code once again, It worth mentioning that `retain_mut` panic and drop handling behavior of smallvec and rust library are different.
It would be nice if for version 2, smallvec start behaving like rust in case of panic.

_Originally posted by @fereidani in https://github.com/servo/rust-smallvec/issues/443#issuecomment-5402692308_

LunaMeerkats:
## Summary

- make `SmallVec::retain_mut` reuse `ExtractIf`'s existing panic backshift guard
- preserve unprocessed elements when the predicate panics
- stop visiting later elements and repair the vector when dropping a rejected element panics
- cover both inline and spilled storage for predicate and destructor panics

The issue claim mentioned a direct port of `Vec::retain_mut`'s guard. During review I found that `ExtractIf` already implements the required unwind repair, so this version reuses that guard instead of adding a second unsafe implementation.

## Reproduction and validation

All four new tests fail independently on pinned base `b675995e70144f03fd5c5e0ce56d3f98115d0950` and pass with this change.

- `cargo build --verbose`
- `cargo test --verbose` — 68 unit + 1 integration + 6 doctests passed
- `cargo test --verbose --features serde` — 69 unit + 1 integration + 6 doctests passed
- `cargo test --verbose --features malloc_size_of` — 68 unit + 1 integration + 6 doctests passed
- `cargo check --verbose --no-default-features`
- `cargo fmt --all -- --check`
- `git diff HEAD^ --check`

A strict all-target Clippy run is blocked by warnings/lints also reproduced on the pinned base; a library Clippy run passes when allowing those exact baseline lints. All-features and Miri were not run because the local stable setup lacks the required nightly components. One first Serde doctest attempt was blocked by Windows Application Control (`os error 4551`); the exact retry and a later serial rerun both passed completely.

Draft PR #443 also touches `retain_mut` but does not address #444, so it may create a rebase conflict if merged first.

Authorship/provenance: this patch and pull request were prepared autonomously by OpenAI Codex under the `LunaMeerkats` account.

Fixes #444

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1610,21 +1610,7 @@ impl<T, const N: usize> SmallVec<T, N> {
 
     #[inline]
     pub fn retain_mut<F: FnMut(&mut T) -> bool>(&mut self, mut f: F) {
-        let mut del = 0;
-        let len = self.len();
-        let ptr = self.as_mut_ptr();
-        for i in 0..len {
-            // SAFETY: all the pointers are in bounds
-            // `i - del` never overflows since `del <= i` is a maintained invariant
-            unsafe {
-                if !f(&mut *ptr.add(i)) {
-                    del += 1;
-                } else if del > 0 {
-                    core::ptr::swap(ptr.add(i), ptr.add(i - del));
-                }
-            }
-        }
-        self.truncate(len - del);
+        self.extract_if(.., |element| !f(element)).for_each(drop);
     }
 
     #[inline]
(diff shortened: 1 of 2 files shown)


alejandro-vaz:
AI contributions are not allowed in any @servo repository as indicated on the contributing guidelines

this PR will subsequently be closed

PR #449, opened 2026-08-26
Fix Stacked Borrows UB when splicing an inline SmallVec

jaideeppyne:
`SmallVec::splice` triggers undefined behavior (a Stacked Borrows aliasing violation, caught by `cargo miri test`) when the vector is stored inline:

```rust
let mut v: SmallVec<i32, 16> = smallvec![1, 2, 3, 4, 5];   // stays inline
let removed: Vec<i32> = v.splice(1..3, [10, 11, 12]).collect();
```

`Drain::fill` builds a `&mut [T]` into the inline buffer and then calls `vec.set_len(..)` inside the fill loop. Since the inline buffer lives inside the `SmallVec`, a `&mut self` method reborrows the whole struct and invalidates that slice (and the `place` reference derived from it), so the next loop iteration retags from a dead tag:

```
error: Undefined Behavior: trying to retag from <128112> for Unique permission at
alloc41362[0x10], but that tag does not exist in the borrow stack for this location
help: <128112> was created by a SharedReadWrite retag at offsets [0xc..0x14]
help: <128112> was later invalidated at offsets [0x0..0x28] by a Unique function-entry retag
```

`Drain::move_tail` has the same issue: it takes a `*const` via `vec.as_ptr()` and then reborrows the whole `SmallVec` again through `vec.as_mut_ptr()`, invalidating the first pointer before `ptr::copy` uses it. Both only manifest while inline, which is why the existing `splice` test (`SmallVec<u8, 1>`, always spilled) never caught it.

Fixed the way std's `Drain::fill` avoids it: update the length by writing the `len` field directly instead of calling a method, and derive the copy's source and destination from a single base pointer. Adds `tests::splice_inline`, which fails under `cargo miri test` before this change and passes after. The full suite stays green under both Stacked Borrows and Tree Borrows.

One note for your judgement: this is a **Stacked Borrows** violation — it passes under Tree Borrows, which tracks that `set_len` only writes the `len` field. Both models are still experimental, but this is UB under the model Miri runs by default and that `scripts/run_miri.sh` uses in CI, it's reachable from entirely safe code, and holding a reference into a struct-embedded buffer across a `&mut self` reborrow is the pattern std deliberately avoids. The fix costs nothing either way.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -492,6 +492,9 @@ impl<T, const N: usize> Drain<'_, T, N> {
         let vec = unsafe { self.vec.as_mut() };
         let range_start = vec.len();
         let range_end = self.tail_start;
+        // Whether the buffer lives on the heap does not change while filling, so
+        // read it once, before `range_slice` is created.
+        let on_heap = vec.len.on_heap();
         let range_slice = unsafe {
             core::slice::from_raw_parts_mut(
                 vec.as_mut_ptr().add(range_start),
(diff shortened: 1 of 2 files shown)


alejandro-vaz:
AI contributions are not allowed in any @servo repository as indicated on the contributing guidelines

this PR will subsequently be closed

jdm:
@alejandro-vaz What signal did you use to decide this was an AI contribution?

alejandro-vaz:
@jdm em-dashes, perfect syntax, the generic summary-implementation notes, and a comment in #444 saying it was literally a Codex agent

and if you look at the code, agents tend to add tests for everything, helper functions where there should be none, and increase LOC way too much for what was necessary

look at each individual one

all these contributions started appearing when I labeled issues as "good first issue" and "help wanted" which is amazing for attracting contributors and people with agents

correct me if I'm wrong but this looks really AI-made

PR #456, opened 2026-08-27
#396 recovery

fereidani:
This recover #396 changes which were rollback because of a MSRV problem with criterion.

* Reduces iterations of test under miri so it is runnable in reasonable time.
* Optimizes pop implementation
* Optimizes retain_mut implementation
* Adds criterion(0.4) benchmarks

I removed reserve optimizations as I am more experienced now and I can do better than that.
I'll send that and fix for #444 in separate PRs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,16 +3,358 @@
 version = 4
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


alejandro-vaz (approved):

alejandro-vaz:
> I removed reserve optimizations as I am more experienced now and I can do better than that.
> I'll send that and fix for https://github.com/servo/rust-smallvec/issues/444 in separate PRs.

sure, thanks

PR #460, opened 2026-08-27
ci: build docs with all features on nightly

### LINKED ISSUE #459: add CI doc build
there should be a CI nightly run that checks docs are built successfully with

```sh
cargo doc --all-features
```

so documentation doesn't break unexpectedly without CI error

this avoids cases like [v1.15.2 not having docs.rs public documentation](https://docs.rs/crate/smallvec/latest)

ITSMERNB:
## Summary
- build documentation with all features in the existing nightly toolchain job
- leave stable, beta, and MSRV jobs unchanged

## Validation
- `cargo +nightly doc --all-features`
- `git diff --check`

Closes #459

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -57,6 +57,10 @@ jobs:
         if: matrix.toolchain == 'nightly'
         run: cargo test --verbose --all-features
 
+      - name: Cargo doc all features
+        if: matrix.toolchain == 'nightly'
+        run: cargo doc --all-features
+
       - name: Cargo bench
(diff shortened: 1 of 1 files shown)


alejandro-vaz (approved):
I can't determine if this was made with AI, which goes against our contributing guidelines

please don't make use of it in the repositories of @servo

thanks for contributing

ITSMERNB:
Thanks for flagging this. Understood — I’ll respect Servo’s contribution policy and won’t submit AI-assisted patches to @servo repositories going forward. I appreciate the clarification and the review.

PR #461, opened 2026-08-27
Add an "allocator-api2" feature

### LINKED ISSUE #55: Support parametric allocators
Add support for passing a custom allocator with the new [allocator API](https://github.com/rust-lang/rust/issues/32838).

TDecking:
Second attempt at creating custom allocators for `SmallVec`.

This one uses the `allocator-api2` crate (enabled by the feature with the same name) to provide provide access
to allocators in stable Rust. If the feature is disabled, the allocator api provided by nightly Rust is used instead.

Closes #55.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "allocator-api2"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c880a97d28a3681c0267bd29cff89621202715b065127cd445fa0f0fe0aa2880"
+dependencies = [
+ "serde_core",
+]
+
+[[package]]
 name = "anes"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -500,6 +509,7 @@ checksum = "0c790de23124f9ab44544d7ac05d60440adc586479ce501c1d6d7da3cd8c9cf5"
 name = "smallvec"
 version = "2.0.0-alpha.12"
 dependencies = [
+ "allocator-api2",
  "bytes",
  "criterion",
  "malloc_size_of",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,16 +14,19 @@ documentation = "https://docs.rs/smallvec/"
 exclude = [".gitignore", "tests/", "scripts/", "fuzz/", "benches/", ".github/"]
 
 [features]
-std = []
+default = ["allocator-api2"]
+std = ["allocator-api2?/std"]
 specialization = []
 may_dangle = []
-serde = ["dep:serde_core"]
+serde = ["dep:serde_core", "allocator-api2?/serde"]
 internals = []
+allocator-api2 = ["dep:allocator-api2"]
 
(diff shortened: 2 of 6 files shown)


alejandro-vaz:
CI is crying

alejandro-vaz:
this looks good

it's +500 LOC more but it's expected given the feature it's adding

is it now ready for review??

alejandro-vaz:
I'm not sure how this will interact with your original PR

on one side we have the nightly allocator API

on the other side we have the allocator-api2, which is a reflection of the nightly API ported to stable

this is more complex than I initially thought it to be

we basically have these cases:
- stable + no allocator
- stable + allocator-api2
- nightly + no allocator
- nightly + allocator-api
- nightly + allocator-api2

but since api2 is just api re-exported, this is simply a normal feature that works when enabled independently of the rust build

and since api2 re-exports api, we can simply never link against the unstable allocator-api and use allocator-api2. that will also help us by removing a potential version mismatch between allocator-api2 and the current nightly implementation

regarding the implementations, I think we should go dead-simple: use an allocator generic internally for `SmallVec` and `RawSmallVec` (not sure but I think it'd be necessary). and if the feature is disabled, then we export `SmallVec` with the last type argument being `Global`

let me know your opinion on this

TDecking:
The first PR can be closed now. It was an attempt to get the "stable + no allocator" and "nightly + allocator-api" under one hood without any dependency. It worked, but it was also ugly. I've opted to create a new PR for supporting `allocator-api2` since it would mean a complete rewrite anyway and also as a refresher; I didn't look at the library for several months.

Of all the cases you've listed, "stable + no allocator" and "nightly + no allocator" should be of no concern, because the allocator type variable defaults to `Global` if not specified. `SmallVec<T, N>` can be used either way. This should correspond to the "dead-simple" approach you suggested.

`allocator-api2` is unfortunately not a simple re-export `allocator-api` except for the stable parts (`Layout`, `AllocError`).
Any change in the nightly Rust API (this PR was hit by one) will mean that this crate needs to support two APIs until
`allocator-api2` updates. This is hopefully a surmountable problem.
The way you framed it suggest that you propose to support only `allocator-api2`. Is that right?

Something that needs to be mentioned is that https://github.com/rust-lang/rust/pull/156882 seems to landing soon (?),
which would open up another option: Unconditionally support custom allocators at the expense of supporting only one Rust version. `allocator-api2` promises to just re-rexport std allocation methods once they are stable, the dependency can then be removed.

alejandro-vaz:
this is becoming insanely complicated

out of all options we have, knowing what you just said that it's almost stabilizing, I'm simply say it's better to wait and implement it properly for the few versions that have it stable

the nightly feature + allocator-api2 setup is just too cumbersome, and having allocator-api2 in the dependency graph certainly isn't nice, it's kind of a smell

we still have a few more months to figure this out, if by the time beta.1 release date approaches (12th October) we see it better, then we implement it in alpha, otherwise we'll have it in beta

we could also try to postpone a bit the releases. say, 9th November beta.1 and 15th December rc.1 or something like that

alejandro-vaz:
let's move discussion to #55, we have to figure out alternatives and what can we do fully in the first place

PR #500, opened 2026-08-29
Add try_with_capacity fallible constructor

### LINKED ISSUE #416: Implement `try_with_capacity`
It would be nice to have an analogue to the nightly [`Vec::try_with_capacity`](https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.try_with_capacity) in SmallVec.

Jorge-Polanco-Roque:
Closes #416.

Adds `try_with_capacity`, the fallible analogue to `with_capacity`. It mirrors the existing `try_reserve` path: uses `try_grow` and returns `Result<Self, CollectionAllocErr>`.

```rust
pub fn try_with_capacity(capacity: usize) -> Result<Self, CollectionAllocErr>
```

I kept it ungated, matching `try_reserve`/`try_grow` (also ungated). You mentioned it might live behind a feature gate — happy to move it if you'd prefer; just let me know which feature.

Test added in `tests/main.rs` covering the inline and spilled cases.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -849,6 +849,19 @@ impl<T, const N: usize> SmallVec<T, N> {
         this
     }
 
+    /// Constructs a new, empty `SmallVec` with at least the specified capacity,
+    /// returning an error if the allocation fails.
+    ///
+    /// This is the fallible version of [`with_capacity`](Self::with_capacity).
+    #[inline]
+    pub fn try_with_capacity(capacity: usize) -> Result<Self, CollectionAllocErr> {
+        let mut this = Self::new();
+        if capacity > Self::inline_size() {
+            this.try_grow(capacity)?;
+        }
+        Ok(this)
+    }
+
     #[inline]
     pub const fn from_buf<const S: usize>(elements: [T; S]) -> Self {
         const {
--- a/tests/main.rs
+++ b/tests/main.rs
@@ -126,6 +126,19 @@ fn test_with_capacity() {
 }
 
 #[test]
+fn test_try_with_capacity() {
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -849,6 +849,19 @@ impl<T, const N: usize> SmallVec<T, N> {
...
+    /// This is the fallible version of [`with_capacity`](Self::with_capacity).
+    #[inline]
+    pub fn try_with_capacity(capacity: usize) -> Result<Self, CollectionAllocErr> {
+        let mut this = Self::new();
+        if capacity > Self::inline_size() {
+            this.try_grow(capacity)?;
+        }
+        Ok(this)
+    }
+
```
alejandro-vaz, line 864: this adds unnecessary overhead

the point of `try_with_capacity` is that it simplifies `try_grow` by realizing what its initial state is

for example, here `try_with_capacity` will call `try_grow` which will check things like "is the instance spilled??" "what is its length??"

all that overhead can be removed

alejandro-vaz (changes requested):

alejandro-vaz:
also, AI contributions are not allowed in any @servo repository, I'm closing this PR

please read it

src/lib.rs
bolshoytoster: This is just the mirror of the implementation for `with_capacity`. Should I see if that could benefit from using this information as well?

bolshoytoster: I've benchmarked it, and it seems `with_capacity` can be improved by ~12% by changing it to:
```rs
let mut this = Self::new();
if capacity > Self::inline_size() {
    infallible(unsafe { this.raw.try_grow_raw(0, capacity) });

    unsafe { this.set_on_heap() };
}
this
```

PR #445, opened 2026-08-25
Deprecate smallvec! and smallvec_inline!

bolshoytoster:
See #432 for details.

Issue #491, opened 2026-08-29
add full feature list to `./README.md`


alejandro-vaz:
add all the features in `Cargo.toml` with a simple explanation to the README (one paragraph each, don't go crazy)

this allows us in the future to remove the massive doc-comment from `lib.rs`

> there is already a simple explanation in the first 50 lines of `lib.rs`, but not all features there might exist anymore, and there might be new ones as well

current features are:
- `std`
- `specialization`
- `may_dangle`
- `serde`
- `internals`
- `bytes`
- `malloc_size_of`

PR #520, opened 2026-08-30
docs: stop using deprecated smallvec! macro in README example

### LINKED ISSUE #518: update `README.md` example to not use the `smallvec!` macro
the `smallvec!` and `smallvec_inline!` macros were deprecated in favor of `SmallVec::from` and `SmallVec::from_buf` in #445 

thus, we shouldn't be showing it off as an example in the README

simply replace that

this is how the example currently looks like

```rust
    
// This SmallVec can hold up to 4 items on the stack:
let mut v: SmallVec<i32, 4> = smallvec![1, 2, 3, 4];

// It will automatically move its contents to the heap if
// contains more than four items:
v.push(5);

// SmallVec points to a slice, so you can use normal slice
// indexing and other methods to access its contents:
v[0] = v[1] + v[2];
v.sort();
```

painbaba:
Fixes #518.

## What

The README example still showcased the `smallvec!` macro, which has been deprecated since `2.0.0-alpha.13` (see #445, deprecation notes in `src/lib.rs`: `use SmallVec::from instead`). This PR updates the example to the blessed API:

```rust
let mut v: SmallVec<i32, 4> = SmallVec::from([1, 2, 3, 4]);
```

and drops the macro from the `use` statement. It also removes a line of trailing whitespace that was on the old import line.

## Why

New users copy the README first. Advertising a deprecated macro in the flagship example pushes them toward code that emits deprecation warnings on day one, contradicting the crate's own guidance from #445.

## Verification

- `SmallVec::from([T; M])` is provided by the unconditional `impl<T, const N: usize, const M: usize> From<[T; M]> for SmallVec<T, N>` (`src/lib.rs`, line ~2726) - no feature gate, works in `no_std` core contexts just like the macro did.
- The README block is not a doctest (not `include_str!`-ed into lib.rs), so no test churn; example semantics are unchanged (4 inline elements, spill on `push(5)`).

## Note / possible follow-up

`src/lib.rs` line ~1903 still uses `smallvec![1, 2, 3]` inside a doctested example on `as_mut_ptr` (that one IS compiled by CI). Left out of this PR to keep it docs-only and scoped to #518 - happy to do a follow-up if useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,10 @@
 ## Example
 
 ```rust
-use smallvec::{SmallVec, smallvec};
-    
+use smallvec::SmallVec;
+
 // This SmallVec can hold up to 4 items on the stack:
-let mut v: SmallVec<i32, 4> = smallvec![1, 2, 3, 4];
+let mut v: SmallVec<i32, 4> = SmallVec::from([1, 2, 3, 4]);
 
 // It will automatically move its contents to the heap if
 // contains more than four items:


alejandro-vaz:
is this ready for review

alejandro-vaz:
it looks like to me

alejandro-vaz (approved):

alejandro-vaz:
AI contributions are not allowed in any @servo repository

thanks for contributing

alejandro-vaz:
I'm not sure if you're a human being or not, I've received way too many AI PRs recently

if you want you can try #491, it's also docs

thank you for your contribution

PR #521, opened 2026-08-31
 added expectations to #[should_panic] tests

### LINKED ISSUE #502: add `#[should_panic]` attribute expect options on `tests/*.rs`
currently we use a ton of `#[should_panic]` attributes on our tests

most of them don't include an expect inside them, but some do. `tests/bytes.rs:84`:

```rust
#[should_panic(expected = "advance out of bounds: the len is 8 but advancing by 12")]
fn // ...
```

to make it consistent all `#[should_panic]` attributes on `tests/*.rs` should have an expect reason added to them

some of the exact errors are not deterministic, but the error type is, and since expect matches a substring it's easier to maintain if we just keep the error name without details

good example: "advance out of bounds"

bad examples:
- the full error with details: "advance out of bounds: the len is 8 but advancing by 12"
- no expect, just bare `#[should_panic]`
- anything that does not match the error name

prakrititz:
Closes #502

Apologies, I missed linking the issue when I first opened this.

Several tests in tests/main.rs and tests/bytes.rs used should_panic without an expected string. This PR adds deterministic expected substrings to all bare should_panic attributes.


alejandro-vaz:
```
failures:
    test_put_int_le_nbytes_overflow
    test_put_int_nbytes_overflow
```

alejandro-vaz (approved):

PR #534, opened 2026-08-31
fix(deps): pin dependency versions in Cargo.toml

### LINKED ISSUE #532: lock all dependency versions to full semver
to avoid possible issues with updating from patches automatically, lock all dependency versions in `Cargo.toml` to their full semver 

that means `=1.2.3` for example

```toml
[dependencies]
arbitrary = { version = "1.4", optional = true, default-features = false }
borsh = { version = "1.8", optional = true, features = ["derive", "unstable__schema"], default-features = false }
bytes = { version = "1.12", optional = true, default-features = false }
defmt = { version = "1.1", optional = true, default-features = false}
serde_core = { version = "1.0", optional = true, default-features = false }
malloc_size_of = { version = "0.1", optional = true, default-features = false }

[dev-dependencies]
serde_test = "1.0"
criterion = "0.8"
```

simply specify version numbers

Dazzanova:
## Summary
- Pin all dependency versions in `Cargo.toml` to their exact versions from `Cargo.lock`.

## Problem
- `Cargo.toml` uses loose dependency versions, which can allow automatic patch updates and cause unexpected issues.

## Solution
- Updated all dependencies to use exact `=x.y.z` versions matching `Cargo.lock`.

## Related
Closes #532

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,16 +20,16 @@ internals = []
 borsh = ["dep:borsh"]
 
 [dependencies]
-arbitrary = { version = "1.4", optional = true, default-features = false }
-borsh = { version = "1.8", optional = true, features = ["derive", "unstable__schema"], default-features = false }
-bytes = { version = "1.12", optional = true, default-features = false }
-defmt = { version = "1.1", optional = true, default-features = false}
-serde_core = { version = "1.0", optional = true, default-features = false }
-malloc_size_of = { version = "0.1", optional = true, default-features = false }
+arbitrary = { version = "=1.4.2", optional = true, default-features = false }
+borsh = { version = "=1.8.1", optional = true, features = ["derive", "unstable__schema"], default-features = false }
+bytes = { version = "=1.12.1", optional = true, default-features = false }
+defmt = { version = "=1.1.1", optional = true, default-features = false }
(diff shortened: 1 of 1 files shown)


alejandro-vaz (approved):

alejandro-vaz:
thanks for contributing

jdm:
What is the motivation here? This is an unusual practice.

alejandro-vaz:
to make sure that dependency requirements don't change between patch versions

it seems that's what happened to `criterion` that caused #396 to not be able to compile and had to be reverted with #430 before we had to finally get back the PR with criterion 0.4

technically 0.7 would have sufficed, but it seemed a version requiring rustc 1.86+ was published as a patch or something along those lines, so after the 0.7 download (which required 1.80 at publish time) some dep updated it's version of `clap_lex` that required 1.86 invalidating 1.80 for the published version

jdm:
I don't think it's it's a good idea to use exact version pinning like this in a library crate, because any crate that includes smallvec anywhere in its dependency tree will no longer be able to independently upgrade (or downgrade) any crate that smallvec depends upon.

jdm:
Given how common it is for other projects to rely on serde_core and bytes, I think we should revert this change. It's going to confuse and annoy other projects.

PR #536, opened 2026-09-01
perf: reduce binary size by optimizing panic formatting

fereidani:
This PR reduces binary size of users by avoiding inlining of rarely executed panic formatting code, methodology is exactly the same as what is used in rust standard library.

`#[cold]` hint is also expected to improve the performance slightly by marking panic path unlikely for the compiler.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -169,6 +169,17 @@ const fn is_zst<T>() -> bool {
 /// and thus cannot be used yet.
 fn slice_range<R>(range: R, bounds: core::ops::RangeTo<usize>) -> core::ops::Range<usize>
 where R: core::ops::RangeBounds<usize> {
+    #[cold]
+    #[inline(never)]
+    #[track_caller]
+    fn assert_failed(start: usize, end: usize, len: usize) -> ! {
+        if start > end {
+            panic!("slice index starts at {start} but ends at {end}");
+        } else {
+            panic!("range end index {end} out of range for slice of length {len}");
+        }
+    }
+
     let len = bounds.end;
 
     let start = match range.start_bound() {
@@ -187,11 +198,8 @@ where R: core::ops::RangeBounds<usize> {
         core::ops::Bound::Unbounded => len
     };
 
-    if start > end {
-        panic!("slice index starts at {start} but ends at {end}");
-    }
-    if end > len {
-        panic!("range end index {end} out of range for slice of length {len}");
+    if start > end || end > len {
+        assert_failed(start, end, len);
     }
 
     core::ops::Range {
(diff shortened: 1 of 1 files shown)


alejandro-vaz:
what improvements do the benchmarks show??

I'm a bit afraid of these types of changes because they bloat the codebase a lot

is there any way we can extract all those `assert_failed` into a common helper with a generic message or an enum dispatch??

fereidani:
These changes are hard to show in micro-benchmarks, as they are small enough that cpu branch-prediction learns the unlikely path by itself, so I didn't try.
These usually help improving CPU cache hit performance in real programs by reducing instruction-pollution. 

> is there any way we can extract all those assert_failed into a common helper with a generic message or an enum dispatch??

The cost of that will be that passing of the enum itself and its encapsulation increasing the call instructions and therefore increasing binary-size/cache-pollution. I won't suggest it although it would be cleaner.

I would suggest a 3rd reviewer comment on this, as I'm not sure why this hasn't implemented or suggested by now.

alejandro-vaz:
@TDecking you've worked a lot on v2 internals, what are your thoughts

it makes sense to me, though I wonder if it could be optimized at compile time

```rust

assert_failed(
    /* whatever condition here, a bool */, 
    const { Enum::RetainMut } /* whatever the specific name */
)

#[inline(always)]
fn assert_failed(condition: bool, error_type: Enum) -> () {
    match error_type {/* the proper arms here */}
}
```

since `error_type` would be known at compile time, and the function is always inlined and it is just a match, I think it would be fully inlined at compile time and we'll end up with the same codegen

TDecking:
As the original comment said, this is also done inside the standard library.
I think that following standard library pratices is always a good idea.

I looked into one instance (`core::slice::range`). One thing they do that's not done here is checking for `panic = "immediate-abort"`. They inline these panic functions in the immediate abort case, this PR currently doesn't do that.

As for codebase bloat, I'm not sure if it's avoidable. `std::Vec` is a conceptionally simpler type that doesn't need to consider features, and yet has more lines of code when compared to `smallvec`. The best thing that can be done is to keep code quality high, you are thankfully ensuring that.

As for your proposed optimization, I don't think it's a good idea. The oulined functions are inside functions with an inline attribute. This is why outlining is a binary size optimization, because it prevents inlining from repeating panic-related codegen. `assert_failed` also has an inline attribute, so the panic calls over there must be outlined aswell if the binary size optimization is to be kept. The result would be the following:

```Rust
#[inline(always)]
fn assert_failed(error: ErrorType) -> ! {
    #[inline(never)]
    #[cold]
    fn panic1() { panic!() }
    #[inline(never)]
    #[cold]
    fn panic2() { panic!() }
    ...

    match error {
        ErrorType::Err1 => panic1(),
        ErrorType::Err2 => panic2(),
        ...
    }
}

#[inline]
fn use_case_1() {
    assert_failed(ErrorType::Err1)
}

#[inline]
fn use_case_2() {
    assert_failed(ErrorType::Err2)
}

...
```

I don't see a gain over calling the outlined functions directly. Using `#[inline(never)]` on the function isn't the best idea either, because as mentioned in the earlier comment, codegen gets worse. If various panic calls are to be bundled, then it's probably best to put them all inside a dedicated module.

alejandro-vaz:
I agree with the idea of bundling the panic calls into their own module, we can do that in another PR

that's probably the 80/20 here

I'll open an issue for that

alejandro-vaz (approved):

TDecking:
@fereidani I spotted some additional outlining opportunities:

* `infallible`
* `SmallVec::leak`
* `spec_traits::SpecExtend::spec_extend`
* `spec_traits::SpecFromIterator::spec_from_iter`

fereidani:
@TDecking Thanks I will look into it.

PR #539, opened 2026-09-01
Bump v1 to 1.16.0.

jdm:


alejandro-vaz:
you might need to do `rustup update`, a few days rustfmt changed how it handles whitespaces in comments and some stuff like that

jdm:
To avoid that problem, we really should be performing style checks in stable builds.

alejandro-vaz:
but what stable do we select then for that then

MSRV??

otherwise it's going to keep changing

alejandro-vaz (approved):

jdm:
Published v1.16.0.

Issue #451, opened 2026-08-26
simplify the `TaggedLen` / `set_len` / `capacity` ... model


alejandro-vaz:
the current length/capacity model is becoming hard to maintain and understand leading to strange miri provenance bugs #406 

what we need conceptually is:

```
TaggedLen: Copy -> [onheap:1][length:63]
- default() -> Self
- new(bool, usize) -> Self
- length(self) -> usize
- onheap(self) -> bool
- get(self) -> (bool, usize)
- resize(&mut self, usize) -> ()
- relocate(&mut self, bool) -> ()
- update(&mut self, bool, usize) -> ()
```

and no additional methods like `set_len` on `SmallVec`

currently `set_len` depends on `capacity` its downstream methods, so this is not a simple refactor

the sketch above is just a concept

fereidani:
`set_len` is a unsafe public API and which enables users to implement their own specialized optimized code, are we sure we are going to remove it?

alejandro-vaz:
not removing it fully since it is on vec, but certainly not using it everywhere internally and instead operate with `self.len`

PR #547, opened 2026-09-01
perf: increment/decrement len without reading on heap flag

fereidani:
It's pretty simple and it just bypasses reading the on heap flag by adding 0b10 when type is not ZST.

But I think there is another problem here. set_len is a unsafe operation, in this codebase we can set_len from safe code.
We must follow this by a refactor/fix and making it impossible to set_len from safe API to avoid future mistakes.
Maybe a `UnsafeCell<usize>` solution?

### DIFF
--- a/src/taggedlen.rs
+++ b/src/taggedlen.rs
@@ -54,4 +54,26 @@ impl<T> TaggedLen<T> {
     pub const fn value(self) -> usize {
         if Self::IS_ZST { self.0 } else { self.0 >> 1 }
     }
+
+    /// Returns the same tag with the length increased by one.
+    ///
+    /// This increases the length without rereading the `on heap` flag.
+    #[inline]
+    pub const unsafe fn increment(&mut self) {
+        self.0 += if Self::IS_ZST {
+            1
+        } else {
+            debug_assert!(self.value() + 1 < isize::MAX as usize);
+            0b10
+        }
+    }
+
+    /// Returns the same tag with the length decreased by one.
+    ///
+    /// This decreases the length without rereading the `on heap` flag.
+    #[inline]
+    pub const unsafe fn decrement(&mut self) {
+        debug_assert!(self.value() > 0);
+        self.0 -= if Self::IS_ZST { 1 } else { 0b10 };
+    }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1138,10 +1138,11 @@ impl<T, const N: usize> SmallVec<T, N> {
             // We have to do this so that Miri doesn't report a "Stacked
             // Borrows" rule violation. See PR/406
 
-            let new_len = len + 1;
-            debug_assert!(new_len <= self.capacity());
-            let on_heap = self.len.on_heap();
-            self.len = TaggedLen::new(new_len, on_heap);
+            debug_assert!(len + 1 <= self.capacity());
+            // SAFETY: we have wrote the value to the address already
+            unsafe {
+                self.len.increment();
+            }
         }
 
         // SAFETY: `ptr` is aligned, non-null and points to the element
@@ -1157,8 +1158,11 @@ impl<T, const N: usize> SmallVec<T, N> {
             return None;
         }
         let new_len = len - 1;
-        // SAFETY: new_len < len since len is non-zero
-        unsafe { self.set_len(new_len) };
+        // SAFETY: new_len < len since len is non-zero and
+        // we are returning ownership of the current value.
+        unsafe {
+            self.len.decrement();
+        }
         // SAFETY: this element was initialized and we just gave up ownership of
         // it, so we can give it away
         let value = unsafe { self.as_mut_ptr().add(new_len).read() };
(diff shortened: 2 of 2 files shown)


alejandro-vaz:
yeah this is what #451 is trying to do more or less

would you want to implement it in full??

I do like your increment thing, maybe you can generalize it to `AddAssign for TaggedLen` / `SubAssign`

when a type is not a ZST, any increment of length by `value` is just `value * 2` or `value << 1` and then addition

it would also be helpful (a few notes I took):
- adding the ability to relocate (change the LSB without having to create a `TaggedLen`) `fn relocate<const ONHEAP: bool>(&mut self)`
- not having the `TaggedLen` API use an `on_heap: bool` argument but a const generic (it fully simplifies at compile time). it can be done by adding also relocate

alejandro-vaz:
`set_len` is something that we have to try to remove, see #406 and the weird stuff we had to do

fereidani:
>yeah this is what https://github.com/servo/rust-smallvec/issues/451 is trying to do more or less
would you want to implement it in full??

Yeah sure, I have 3 or 4 optimization related PRs for smallvec that I'm working on, I can do it after those, But don't block it because of me, if someone is interested let them do it.

> I do like your increment thing, maybe you can generalize it to AddAssign for TaggedLen / SubAssign

I prefer const fn, I saw some weird stuff with generic code optimizations that for these low level stuff I prefer to keep it as simple as possible.

> when a type is not a ZST, any increment of length by value is just value * 2 or value << 1 and then addition

What about putting the flag in MSB instead and always doing +1 and -1? we can use CPU overflow flags and implement it better. I need to review the codebase more carefully, I'm not sure that integer overflows are handled correctly.

alejandro-vaz:
I did make a try at storing the flag on the MSB, it's basically the same, except you have to make a mask that only has the MSB on. I do agree that it seems that for repeated length changes it should be faster

I'm pretty sure as well that not all integer overflows are handled correctly. literally our current validation is `debug_assert!` doing all the work and praying that nothing in production hits it in a weird pattern

I don't think fuzzing tests that either

fereidani:
Sorry for the force-push, I missed one decrement case.

Would you like to merge this one or you prefer the full rewrite?

alejandro-vaz:
I prefer it now

alejandro-vaz:
the full rewrite is probably going to take some time, so it's not a good idea to rush it to alpha.13

alpha.14 is for September 10

src/taggedlen.rs
```diff
@@ -54,4 +54,32 @@ impl<T> TaggedLen<T> {
...
+    #[must_use]
+    pub const fn decrement(self) -> Self {
+        debug_assert!(self.value() > 0);
+        if Self::IS_ZST {
+            Self(self.0 - 1, PhantomData)
+        } else {
+            Self(self.0 - 0b10, PhantomData)
+        }
+    }
 }
```
alejandro-vaz, line 79: why not have them taking `&mut self` instead??

fereidani: Nice idea, I also made the functions unsafe.

src/taggedlen.rs
```diff
@@ -54,4 +54,30 @@ impl<T> TaggedLen<T> {
...
+    ///
+    /// This decreases the length without rereading the `on heap` flag.
+    #[inline]
+    pub const unsafe fn decrement(&mut self) {
+        debug_assert!(self.value() > 0);
+        if Self::IS_ZST {
+            *self = Self(self.0 - 1, PhantomData);
+        } else {
+            *self = Self(self.0 - 0b10, PhantomData);
+        }
```
alejandro-vaz: this can be simplified to

```rust
self.0 -= 1 + !Self::IS_ZST as usize
```

it removes a branch and having to create the container

src/taggedlen.rs
```diff
@@ -54,4 +54,30 @@ impl<T> TaggedLen<T> {
...
+    ///
+    /// This increases the length without rereading the `on heap` flag.
+    #[inline]
+    pub const unsafe fn increment(&mut self) {
+        if Self::IS_ZST {
+            *self = Self(self.0 + 1, PhantomData);
+        } else {
+            debug_assert!(self.value() + 1 < isize::MAX as usize);
+            *self = Self(self.0 + 0b10, PhantomData);
+        }
```
alejandro-vaz: this can be simplified to

```rust
self.0 += if Self::IS_ZST {
    1
} else {
    debug_assert!(self.value() + 1 < isize::MAX as usize);
    1 << 1 // or simply 2
}
```

though it probably doesn't make much difference, other than removing the need to create a container

alejandro-vaz (changes requested):

src/taggedlen.rs
fereidani: good idea for self.0 usage but the `!Self::IS_ZST as usize` is unreadable, I only use this trick for branch elimination when it is needed and can't be done by the compiler, in this case compiler can optimize the branch away by itself.

alejandro-vaz (approved):

alejandro-vaz:
whatever borsh serialization/deserialization does, miri spends a bunch of time on it

It needs like 10 minutes

PR #548, opened 2026-09-01
Add rayon support

### LINKED ISSUE #296: Is this library support parallel iterator?
Or plans to support this feature？

bolshoytoster:
Closes #296.

This is largely based on [`Vec`'s implementation](https://github.com/rayon-rs/rayon/blob/ee0a00bdb1ab039e178a215ad5712fb7fa58e58f/src/vec.rs#L188-L294).

I haven't added any tests yet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,10 @@ name = "bytes"
 required-features = ["bytes"]
 
 [[test]]
+name = "rayon"
+required-features = ["rayon"]
+
+[[test]]
 name = "serde"
 required-features = ["serde"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ arbitrary = { version = "1.4", optional = true, default-features = false }
 borsh = { version = "1.8", optional = true, features = ["derive", "unstable__schema"], default-features = false }
 bytes = { version = "1.12", optional = true, default-features = false }
 defmt = { version = "1.1", optional = true, default-features = false }
+rayon = { version = "1", optional = true }
 serde_core = { version = "1.0", optional = true, default-features = false }
 malloc_size_of = { version = "0.1", optional = true, default-features = false }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ arbitrary = { version = "1.4", optional = true, default-features = false }
 borsh = { version = "1.8", optional = true, features = ["derive", "unstable__schema"], default-features = false }
 bytes = { version = "1.12", optional = true, default-features = false }
 defmt = { version = "1.1", optional = true, default-features = false }
-rayon = { version = "1", optional = true }
+rayon = { version = "1.12", optional = true, default-features = false }
 serde_core = { version = "1.0", optional = true, default-features = false }
 malloc_size_of = { version = "0.1", optional = true, default-features = false }
 
--- a/src/rayon.rs
+++ b/src/rayon.rs
@@ -21,7 +21,7 @@ use {
     }
 };
 
-pub(crate) struct SliceDrain<'a, T>(slice::IterMut<'a, T>);
+struct SliceDrain<'a, T>(slice::IterMut<'a, T>);
 
 impl<T> Iterator for SliceDrain<'_, T> {
     type Item = T;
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -24,6 +24,7 @@ arbitrary = { version = "1.4", optional = true, default-features = false }
 borsh = { version = "1.8", optional = true, features = ["derive", "unstable__schema"], default-features = false }
 bytes = { version = "1.12", optional = true, default-features = false }
 defmt = { version = "1.1", optional = true, default-features = false }
+rayon = { version = "1", optional = true }
```
alejandro-vaz: specifying minor would avoid `rayon` receiving a breaking change minor update and affecting us

src/rayon.rs
```diff
@@ -0,0 +1,95 @@
...
+        iter::plumbing::{
+            Producer,
+            UnindexedConsumer,
+            bridge_producer_consumer
+        },
+        prelude::ParallelIterator
+    }
+};
+
+pub(crate) struct SliceDrain<'a, T>(slice::IterMut<'a, T>);
```
alejandro-vaz: why make it public??

it can't be accessed from anywhere either

alejandro-vaz (changes requested):

alejandro-vaz:
looks good, CI complains, make sure you have the latest nightly with `rustup update`

a few days they updated how rustfmt handles whitespace in comments

there's already an issue to pin rustfmt to MSRV but that would introduce conflict anyway, so I'm not sure what will we do

alejandro-vaz:
if you want to add tests in another PR it's okay, leave an issue and then do the implementation if you want

src/rayon.rs
bolshoytoster: Whoops, I was just copying code from rayon.

Cargo.toml
jdm: That's not how that works. Specifying a minor version just ensures that the resolved version is never lower than the specified major.minor, and not greater than major.

alejandro-vaz: is it?? I thought it was like in NPM

alejandro-vaz: we should use `rayon-core` now that I think about it, it should have the core traits as well

jdm: A cargo.lock is what will ensure that the versions we test on CI won't change.

alejandro-vaz: I'm not sure how the lockfile interacts with dependabot and us adding/removing dependencies and sometimes the lockfile being partially regenerated

bolshoytoster: > we should use `rayon-core` now that I think about it, it should have the core traits as well

Can we? I don't see any traits in `rayon-core`.

alejandro-vaz: my mistake, it's true, they aren't on `rayon-core`

alejandro-vaz:
I don't like that miri at all

I'll give it a look tomorrow

bolshoytoster:
I'm looking into it, I think [this commit from crossbeam](https://github.com/crossbeam-rs/crossbeam/commit/02fb08a6c608c43a272214c06c1598f043571dd6) might fix it, but that's not in any release.

alejandro-vaz:
crossbeam hasn't been updated in two years

alejandro-vaz:
is there any way we can restructure our implementation not to trigger crossbeam??

bolshoytoster:
I don't think it's an issue with our implementation. Changing `SmallVec<i32, 2>` to just `Vec<i32>` causes the same miri error :(

Does rayon do miri tests? Presumably they would also have to work around this.

alejandro-vaz:
it does not look like rayon has miri tests

alejandro-vaz:
all we can do is either somehow exempt it from miri, or refactor our implementation (which is not at fault) in a way that does not trigger whatever crossbeam is afraid of

bolshoytoster:
I've found a different miri error in rayon, from the code `(0..4).into_par_iter().collect::<Vec<i32>>()`. Maybe I should open an issue there.

We could just remove the test, or maybe put it under `#[cfg(not(miri))]`.

alejandro-vaz:
yeah definitely, open an issue

`#[cfg(not(miri))]` will keep us sane for now

tests/rayon.rs
```diff
@@ -0,0 +1,20 @@
+#[cfg(not(miri))]
+#[test]
+fn rayon() {
+    use {
+        rayon::{
+            iter::ParallelIterator,
+            prelude::ParallelSlice
+        },
+        smallvec::SmallVec
+    };
```
alejandro-vaz: make this a module-level import instead

for consistency with all other tests

Cargo.toml
alejandro-vaz: we should still probably specify minor for consistency and disable default features

alejandro-vaz (changes requested):

tests/rayon.rs
bolshoytoster: I put it there to shut up the unused import warnings on miri, I'll just add a second `#[cfg()]`.

bolshoytoster:
> and disable default features

I don't think rayon has any default features. Should I still disable them for visual consistency?

alejandro-vaz:
I'd lean towards disabling them, though it doesn't probably matter much

alejandro-vaz (approved):

alejandro-vaz:
looks good, great feature

PR #549, opened 2026-09-01
clean artifacts before benchmarking

### LINKED ISSUE #541: change CI rustfmt to MSRV toolchain instead of nightly
in `.github/workflows/rust.yml` change the style check to use MSRV instead of nightly

current workflow:

```yml
      - name: Style check
        if: matrix.nightly
        run: rustup component add rustfmt && cargo fmt --all --check
```

this removes formatting issues related to `rustfmt` updates such as in #539

### LINKED ISSUE #542: add `cargo clean` right before benching in CI
modify the current `rust.yml` CI workflow so that it cleans built artifacts before benching

this cleans cache which removes any incremental compilation and may lead to more reliable measurements

```yml
      - name: Cargo bench all features
        if: matrix.nightly
        run: cargo bench --verbose --all-features
```

originally proposed by @fereidani in #535

malezjaa:
Fixes #542

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -67,7 +67,7 @@ jobs:
 
       - name: Cargo bench all features
         if: matrix.nightly
-        run: cargo bench --verbose --all-features
+        run: cargo clean && cargo bench --verbose --all-features
 
       - name: miri
         if: matrix.nightly


.github/workflows/rust.yml
```diff
@@ -65,6 +65,10 @@ jobs:
         if: matrix.nightly
         run: cargo doc --all-features --verbose
 
+      - name: Clean built artifacts before benchmarking
+        if: matrix.nightly
+        run: cargo clean
+
       - name: Cargo bench all features
         if: matrix.nightly
         run: cargo bench --verbose --all-features
```
alejandro-vaz: it'd probably be simpler if we just replaced

```yml
run: cargo bench --verbose --all-features
```

for

```yml
run: cargo clean && cargo bench --verbose --all-features
```

malezjaa:
Hmm I guess it can't use MSRV since there are a lot of nightly options

.github/workflows/rust.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           target: ${{ matrix.target }}
 
       - name: Style check
-        if: matrix.nightly
+        if: matrix.name == 'MSRV'
```
alejandro-vaz: I've been thinking about it and no one really uses MSRV and has the exact MSRV rustfmt version installed, so that didn't make much sense

probably stable is a better idea

alejandro-vaz (changes requested):

.github/workflows/rust.yml
alejandro-vaz: if that still gives errors, then we'll just let it be like that

the latest nightly is something that can always be agreed upon even if it is cumbersome

malezjaa: Stable will still give errors, I will keep this pr scope to only clean the artifacts and revert the style check changes.

alejandro-vaz:
looks good

thanks for contributing

alejandro-vaz (approved):

PR #553, opened 2026-09-02
refactor: modularize assertion helpers into assertions.rs (fixes #551)

yunaremaia:
Fixes #551.

Moves all cold `#[cold] #[inline(never)] #[track_caller]` assertion helper functions from `lib.rs` into a dedicated `assertions.rs` module:

- `assert_failed(start, end, len)` in `slice_range` → `assertions::slice_range_failed(start, end, len)`
- `assert_failed(index, len)` in `swap_remove` → `assertions::index_out_of_bounds("swap_remove", index, len)`  
- `assert_failed(index, len)` in `remove` → `assertions::index_out_of_bounds("removal", index, len)`
- `assert_failed(index, len)` in `insert_mut` → `assertions::index_out_of_bounds("insertion", index, len)`

Benefits:
- Centralizes all cold panic-path code in one module
- Reduces code duplication (3 identical nested fn definitions → 1 shared module fn)
- Method name is now included in panic messages for easier debugging
- All 66 existing tests pass

PR #554, opened 2026-09-02
refactor: modularize assertion helpers into assertions.rs (fixes #551)

### LINKED ISSUE #551: modularize `assert_failed` panic calls
some `SmallVec` methods use assertions that rely on a helper function marked as `#[cold]` and never inlined. this is to improve cache locality

take all the helper functions like that, rename them, and put it in an `assertions.rs` module

example in `SmallVec::remove`:

```rust
    #[inline]
    pub fn remove(&mut self, index: usize) -> T {
        #[cold]
        #[inline(never)]
        #[track_caller]
        fn assert_failed(index: usize, len: usize) -> ! {
            panic!("removal index (is {index}) should be < len (is {len})");
        }

        let len = self.len();
        if index >= len {
            assert_failed(index, len);
        }
        let new_len = len - 1;
        unsafe {
            // SAFETY: new_len < len
            self.set_len(new_len);
            let ptr = self.as_mut_ptr();
            let ith = ptr.add(index);
            // This item is initialized since index < len
            let ith_item = ith.read();
            copy(ith.add(1), ith, new_len - index);
            ith_item
        }
    }
```

yunaremaia:
Extracts the four inline `assert_failed` panic helpers from `lib.rs` into a dedicated `assertions` module with descriptive panic messages.

### DIFF
--- a/src/assertions.rs
+++ b/src/assertions.rs
@@ -1,0 +1,37 @@
+// Licensed under the Apache License, Version 0.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Cold assertion helpers for SmallVec methods.
+//!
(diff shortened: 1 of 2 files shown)


alejandro-vaz:
closed for the same reason as #553

Issue #379, opened 2025-05-24
Strange bad performance with [u8; 16] and `union` feature


JonathanWilbur:
Forgive me if I'm making some sort of amateur mistake here. I know when it comes to profiling, subtle things done wrong can completely bias a result, but I ran these two tests with `criterion` and it is saying that the `SmallVec<[u8; 16]>` is about 100x _slower_ than the normal `Vec` at `push()`.

These are my tests:

```rust
fn vec_test() {
    let mut v = Vec::with_capacity(13);
    for i in 0..12 {
        v.push(i);
    }
}

fn smallvec_test() {
    let mut v: SmallVec<[u8; 16]> = SmallVec::new();
    for i in 0..15 {
        v.push(i);
    }
}

fn bench_vec(c: &mut Criterion) {
    let mut group = c.benchmark_group("Vec");
    group.bench_function("Normal", |b| b.iter(|| vec_test()));
    group.bench_function("SmallVec", |b| b.iter(|| smallvec_test()));
    group.finish();
}
```

However, when I change it to `SmallVec<[u8; 15]>`, the smallvec is 3x _faster_ than `Vec` (which is a result that I would expect).

**Why would adding that single additional byte make this so slow?** I am using the `union` feature, so I would think that 16 bytes could fit in `SmallVec` while keeping it the same size as a `Vec`. Is this a known problem? Am I making some obvious mistake?

JonathanWilbur:
Huh, it seems like disabling `union` fixed it (not completely, since the non-`union` implementation is probably just going to be a little slower):

```
Vec/SmallVec            time:   [20.751 ns 20.783 ns 20.823 ns]
                        change: [−72.229% −71.968% −71.731%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low mild
  7 (7.00%) high mild
  1 (1.00%) high severe
```

JonathanWilbur:
Actually, even that appears to be slower than `Vec`, even though it is a lot better than the catastrophic performance when `union` is on.

```
Vec/Normal              time:   [694.64 ps 696.88 ps 699.61 ps]
                        change: [+4.2862% +4.9232% +5.7627%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
[...]

Issue #361, opened 2024-10-18
`spilled()` is very slow


Brogolem35:
I am the author of the [markov_str](https://github.com/Brogolem35/markov_str) library and I trying out the `SmallVec` to speed up the creation of my Markov Chains. But I have encountered this issue while benchmarking with [cargo flamegraph](https://github.com/flamegraph-rs/flamegraph): equvalance checks between `SmallVec`s take too long and almost all the time is spent on `spilled()`.

Example:

MarkovChain type:
```rust
pub struct MarkovChain {
	items: HashMap<SmallVec<[Spur; 4]>, ChainItem>,
	state_size: usize,
	regex: Regex,
	cache: Rodeo,
}
```

benchmarked function:
```rust
	/// Adds text as training data. The tokens will be created with the regex of the MarkovChain.
	pub fn add_text(&mut self, text: &str) {
		let tokens: Vec<Spur> = self
			.regex
			.find_iter(text)
			.map(|t| self.cache.get_or_intern(t.as_str()))
			.collect();

		// vec.windows(0) panics for some reason.
		if tokens.is_empty() {
			return;
		}

		// Creating a preallocated buffer and filling and cleaning it instead of creating a new one every loop is way more efficient.
		let mut prevbuf: SmallVec<[Spur; 4]> = SmallVec::with_capacity(self.state_size);
		for win in tokens.windows(tokens.len().min(self.state_size + 1)) {
			let rel = win.last().unwrap();

			for i in 1..win.len() {
				prevbuf.clear();
				for t in win.iter().rev().skip(1).take(i).rev() {
					prevbuf.push(*t);
				}

				match self.items.raw_entry_mut().from_key(&prevbuf) {
					RawEntryMut::Occupied(mut view) => {
						view.get_mut().add(*rel);
					}
					RawEntryMut::Vacant(view) => {
						view.insert(prevbuf.clone(), ChainItem::new(*rel));
					}
				}
			}
```

Crates used:
```toml
hashbrown = "0.15.0"
lasso = {version = "0.7.3", features = ["ahasher", "inline-more"]}
rand = "0.8.5"
regex = "1.11.0"
smallvec = "1.13.2"
```

Flamegraph output:
![image](https://github.com/user-attachments/assets/8a7181c0-e9e6-4e6a-99a8-839b90b5b95a)

[...]

PR #557, opened 2026-09-02
Speed up push by expanding triple_mut

bolshoytoster:
See #379.

This seems to increase binary size by 32 bytes (on a release build, it's closer to 896 in debug), although it might end up being more depending on how many times `push` is inlined.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1187,15 +1187,28 @@ impl<A: Array> SmallVec<A> {
     #[inline]
     pub fn push(&mut self, value: A::Item) {
         unsafe {
-            let (mut ptr, mut len, cap) = self.triple_mut();
-            if *len == cap {
-                self.reserve_one_unchecked();
-                let (heap_ptr, heap_len) = self.data.heap_mut();
-                ptr = heap_ptr;
-                len = heap_len;
-            }
-            ptr::write(ptr.as_ptr().add(*len), value);
-            *len += 1;
+            if self.spilled() {
+                let (mut ptr, mut len_ptr) = self.data.heap_mut();
+                if *len_ptr == self.capacity {
+                    self.reserve_one_unchecked();
+                    let (heap_ptr, heap_len) = self.data.heap_mut();
+                    ptr = heap_ptr;
+                    len_ptr = heap_len;
+                }
+                ptr::write(ptr.as_ptr().add(*len_ptr), value);
+                *len_ptr += 1;
+            } else {
+                let mut ptr = self.data.inline_mut();
+                let mut len_ptr = &mut self.capacity;
+                if *len_ptr == Self::inline_capacity() {
+                    self.reserve_one_unchecked();
+                    let (heap_ptr, heap_len) = self.data.heap_mut();
+                    ptr = heap_ptr;
+                    len_ptr = heap_len;
+                }
+                ptr::write(ptr.as_ptr().add(*len_ptr), value);
+                *len_ptr += 1;
+            };
         }
     }
 


alejandro-vaz:
32 bytes can be safely ignored as not meaningful

alejandro-vaz:
can you paste the improvements table??

bolshoytoster:
```
Current non-union:
test bench_push                        ... bench:         413.05 ns/iter (+/- 11.63)
test bench_push_small                  ... bench:          52.97 ns/iter (+/- 3.21)
test bench_push_vec                    ... bench:         566.10 ns/iter (+/- 21.99)
test bench_push_vec_small              ... bench:          58.42 ns/iter (+/- 3.26)
test bench_pushpop                     ... bench:         711.49 ns/iter (+/- 26.09)
test bench_pushpop_vec                 ... bench:         292.66 ns/iter (+/- 7.47)

New non-union:
test bench_push                        ... bench:         390.38 ns/iter (+/- 18.83)
test bench_push_small                  ... bench:          39.83 ns/iter (+/- 6.27)
test bench_push_vec                    ... bench:         414.47 ns/iter (+/- 16.19)
test bench_push_vec_small              ... bench:          46.17 ns/iter (+/- 4.24)
test bench_pushpop                     ... bench:         581.87 ns/iter (+/- 104.08)
test bench_pushpop_vec                 ... bench:         233.12 ns/iter (+/- 27.54)

Current union:
test bench_push                        ... bench:         610.56 ns/iter (+/- 12.40)
test bench_push_small                  ... bench:         100.19 ns/iter (+/- 2.04)
test bench_push_vec                    ... bench:         623.76 ns/iter (+/- 25.83)
test bench_push_vec_small              ... bench:          52.79 ns/iter (+/- 5.81)
test bench_pushpop                     ... bench:       1,080.09 ns/iter (+/- 42.82)
test bench_pushpop_vec                 ... bench:         270.46 ns/iter (+/- 11.53)

New union:
test bench_push                        ... bench:         352.68 ns/iter (+/- 12.00)
test bench_push_small                  ... bench:          37.70 ns/iter (+/- 0.74)
test bench_push_vec                    ... bench:         395.48 ns/iter (+/- 27.31)
test bench_push_vec_small              ... bench:          43.04 ns/iter (+/- 2.02)
test bench_pushpop                     ... bench:         659.03 ns/iter (+/- 47.59)
test bench_pushpop_vec                 ... bench:         220.82 ns/iter (+/- 7.93)
```

alejandro-vaz:
improvements are really striking for what the change does

alejandro-vaz:
was this related to #361??

alejandro-vaz (approved):

bolshoytoster:
> was this related to #361??

I still can't reproduce that, but this should improve that anyway.

alejandro-vaz:
I'm not sure I can justify a full patch release from this even though improvements are meaningful 

we should try to avoid releasing two versions within three days at least, as a rule of thumb

this undoubtedly should be published soon

bolshoytoster:
I've just noticed that `insert` has the same pattern:
https://github.com/servo/rust-smallvec/blob/fadb3b8f9d7fd750154aff2a6e17e3afc5ea14f9/src/lib.rs#L1461-L1467

That might also be improvable. I'll look into it.

alejandro-vaz:
is it a case against using `triple` and `triple_mut` altogether or is it just specific to those operations??

if `triple` and `triple_mut` are consistently slower everywhere, we may as well remove them

bolshoytoster:
The problem with this optimization is the potential bloat (and it makes the code more verbose/less maintainable), so it's probably only worth adding it to hot code.

alejandro-vaz:
code can always be refactored, I don't think that's an issue

I haven't started going intra-module to see how to refactor things much, but there is a lot of unnecessary stuff everywhere

if you look at it, it's the same pattern always

you can potentially extract that to a function that's always inlined which takes a `impl FnOnce(/* args of triple */) -> whatever` which is the body of it and the compiler should optimize it the same way

it just looks so surreal that moving that affects codegen so much

bolshoytoster:
I've implemented this for `insert`, but the benchmark results aren't much different. Although, I accidentally found that it does speed up inserting to the end of a `SmallVec` (just a push), so I assume the reason results are the same is because most of the cost is in copying elements, making this optimization insignificant.

Should I submit it anyway, even if the change is unnoticable in most cases, and it makes the code harder to read?

Unrelated: Since `reserve_one_unchecked` is only used in `push` and `insert`, always in this pattern, would it be better to have it return `heap_mut`, i.e.
```rs
fn reserve_one_unchecked(&mut self) -> (NonNull<A::Item>, &mut usize) {
    debug_assert_eq!(self.len(), self.capacity());
    let new_cap = self
        .len()
        .checked_add(1)
        .and_then(usize::checked_next_power_of_two)
        .expect("capacity overflow");
    infallible(self.try_grow(new_cap));
    unsafe { self.data.heap_mut() }
}
```
Then the if blocks that are repeated could just be:
```rs
if *len_ptr == self.capacity {
    (ptr, len_ptr) = self.reserve_one_unchecked();
}
```

I've checked, and this lowers binary size in both debug and release builds.

alejandro-vaz:
I don't think it's a good idea to complicate `insert` if there are no performance gains. it's true that your way is more logically correct and in normal conditions I'd certainly prefer it, but this codebase is already bloated everywhere, right now it's not a good idea

if the private function `reserve_one_unchecked` is only used twice and we are not sure we want its exact behavior, why the hell is it a function then?? just inline copy-paste the code in those two places and adapt it needed. what a clown function

bolshoytoster:
> why the hell is it a function then??

It's worth noting that it's marked `#[cold]` to stop it from being inlined, and it's [part of an optimization](https://github.com/servo/rust-smallvec/commit/1d8f3d7175722c53b04d367e1a3c586daa7b35f5) + using it is more concise than copy-pasting the body everywhere.

alejandro-vaz:
okay, can we do what it does in a simpler way??

it makes sense that removing that function would also put in doubt `reserve` and all others so it's probably not a good idea, you are right

we currently have like four different `reserve` functions that I've seen and #562 wants to add more on v2

since this is v1, I think it's better to shrug our shoulders

if there are performance gains by making it return `heap_mut`, then go ahead

bolshoytoster:
I've benchmarked a few different combinations of changes, these are the ones that weren't eliminated:

| | Debug | Release |
|-|-|-|
| `reserve_one_unchecked` -> `heap_mut` | 184 bytes smaller, 0.1% faster | 64 bytes smaller, 0.88% slower |
| remove `#[cold]`, no -> `heap_mut` | 152 bytes smaller, 0.1% faster | 16 bytes smaller, 0.24% faster |

Note that the changes in speed only apply to a vec that needs to be expanded (I benchmarked adding 16 elements to a `SmallVec<[u8; 1]>`. There is no difference in speed pushing to a vec that never needs to expand.

The second option changes the blocks to
```rs
self.reserve_one_unchecked(); 
(ptr, len_ptr) = self.data.heap_mut(); 
```
since that improves debug builds for some reason, and also looks nicer than the original.

Are either of these worth doing?

alejandro-vaz:
I'm not sure to be honest

is the juice worth the squeeze??

I wouldn't take anything <1% change in speed as proof of anything

I think we probably have other optimizations that we haven't touched that working on them has a higher ROI

use your judgement on whether they are worth it, and I'd simply say have the one of the three alternatives that makes the codebase consistent and easier to understand

if you think one is worth it, open a PR

PR #563, opened 2026-09-02
perf: reduce scope of ptr in retain_mut

fereidani:
I'm surprised but this improves performance and reduces count of assembly instructions.


alejandro-vaz (approved):
it makes sense I guess

instead of fetching the pointer in each iteration it only does so once

fereidani:
I assumed before that compiler can see that pointer is not changing by itself.

PR #573, opened 2026-09-03
Minor style/bloat improvement

bolshoytoster:
Mentioned in #557.

This makes some repeated code more concise and (on debug) lowers bloat from `push` by ~152 bytes (I haven't measured `insert`) and adds a very slight (0.1%) speed improvement to expanding a vec.

There is no effect in release builds.


bolshoytoster:
Nevermind, it looks like the MSRV doesn't support this syntax.

PR #577, opened 2026-09-03
Add a test for the `defmt` feature

### LINKED ISSUE #475: defmt testing
adds proper testing for the feature with corner cases included

only `SmallVec` implements `defmt::Format` as of now

TDecking:
Closes #475.

This test does a lot less than one might expect because `defmt` takes up a lot of responsibility for correctness.

The official documentation (https://defmt.ferrous-systems.com/faq-for-library-authors) recommends to use QEMU to
run a cross compiled binary for testing. I decided not to do that. Instead, I used a crate for connecting `defmt` with the `log` crate, which is IMHO a better apprach.

Because `defmt` is designed for embedded systems, this test won't run on either Windows or with miri. Windows is not supported by `defmt`, and the program tries to access certain linker sections of its binary, which runs into issues with miri's isolation.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -29,6 +29,7 @@ jobs:
             nightly: true
           - name: MSRV
             toolchain: "1.86.0"
+            skip: --skip defmt
           - name: no_std
             toolchain: stable
             target: thumbv7m-none-eabi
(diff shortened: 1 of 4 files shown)


alejandro-vaz:
is it worth it to upgrade to 1.88 just for those kinds of let expressions??

two version = three months, which puts us at a +2.5 month surplus from the original baseline

alejandro-vaz:
don't do it that way, refactor code, it's a test anyway

it's not worth it

alejandro-vaz:
I investigated it a bit, it seems it's way more complex than I initially thought

it looks like a dead end

I guess we'll simply assume it's right and move on

alejandro-vaz:
thanks anyway

PR #578, opened 2026-09-04
refactor: improve taggedlen maintainability and logic

fereidani:
In relation to #451,
This PR refactors taggedlen for better maintainability and slight sub percent performance improvement.

## Highlights
* with `len.parts()` we can stop reading on_heap/spilled flag multiple times.
* ptr construction explicitly requires  providing the on_heap value which avoids constructing invalid pointers and avoids recalculation of on_heap value inside function bodies again. 
* using new const    `MAX_LEN`, `SHIFT` and `TAG` simplifies maintaining the taggedlen.rs.


fereidani:
This also fixes clippy CI failure.

alejandro-vaz (approved):
yeah this makes much more sense than what we had and it's easier to maintain

PR #587, opened 2026-09-06
perf: remove index panic code generation from ExtractIf next

fereidani:
While reviewing panic paths of smallvec I noticed we have a index check panic in `ExtractIf` `next` function.

I checked it with the rust alloc itself and updated the algorithm to remove that mistake.
I believe this mistake was a port of the old rust alloc which got optimized later.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -449,21 +449,19 @@ where F: FnMut(&mut T) -> bool
         unsafe {
             while self.idx < self.end {
                 let i = self.idx;
-                let v = core::slice::from_raw_parts_mut(self.vec.as_mut_ptr(), self.old_len);
-                let drained = (self.pred)(&mut v[i]);
+                // SAFETY: `i < self.end <= self.old_len`
+                let cur = self.vec.as_mut_ptr().add(i);
+                let drained = (self.pred)(&mut *cur);
                 // Update the index *after* the predicate is called. If the
-                // index is updated prior and the predicate
-                // panics, the element at this index would be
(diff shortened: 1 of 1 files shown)


fereidani:
note: The PR code matches the current rust alloc implementation

fereidani:
You can check the panic paths further using `panicgraph -l 8000` on the project. Removing panic paths in bottlenecks results in branchless code and therefore better performance.

alejandro-vaz:
yeah it's those tiny indexes and stuff like that that introduce unnecessary bound checks we already know code fulfills but LLVM can't

you could have gotten your way as well by doing unchecked accesses wouldn't you??

alejandro-vaz (approved):

fereidani:
> yeah it's those tiny indexes and stuff like that that introduce unnecessary bound checks we already know code fulfills but LLVM can't

Exactly, and that we are using unsafe anyway, so there is no need to use safe API and defeat the gain of unsafe.

> you could have gotten your way as well by doing unchecked accesses wouldn't you??

Can you please give me an example? I'm not sure about what you mean.

alejandro-vaz:
```rust
unsafe { v.get_unchecked_mut(i) }
```

it's basically the same except that it doesn't do bound checks

fereidani:
Yeah but I think that would be ugly, and I would have to watch for double drops, I'll prefer pointer API for these kind of stuff.

alejandro-vaz:
more of the same

PR #588, opened 2026-09-06
Refactor `PartialEq` implementations

TDecking:
* Introduces the same macro `alloc` uses for implementing `PartialEq` for its linear data structures, removing a bit of boilerplate.
* Some implementations of `PartialEq` that `Vec` has but `SmallVec` has not are implemented,
  making it easer to use `SmallVec` as a drop-in replacement for `Vec`.
* Adds `PartialEq` implementations for `Vec`, such that comparisons like `smallvec == vec` work
  without having to call `as_slice` or indexing.


alejandro-vaz (approved):
I really disagree with standard library macro bloat instead of writing things, but this is rust so I better stfu

PR #592, opened 2026-09-08
Add custom allocator support

### LINKED ISSUE #55: Support parametric allocators

bolshoytoster:
Closes #55.

This doesn't add support in the following APIs:
From<anything>
from_elem
Default
from_buf
from_buf_and_len
from_buf_and_len_unchecked
from_vec
from_raw_parts
splice
from_slice_copy
FromIterator

`new`, `try_with_capacity` and `with_capacity` now have `_in` counterparts, to match the standard library.

`PartialEq` does not support comparing two `SmallVec`s with different allocators (this is different from std `Vec`). This shouldn't matter too much, since you can just compare them as slices instead.

You also can't `append` two `SmallVec`s with different allocators, this matches std `Vec`.


bolshoytoster:
I'm not really sure how to read the fuzz failure. It looks like one of the tests times out, but I can't seem to install honggfuzz locally to test (`TRUE` being undefined in a c file)

fereidani:
Try to rerun the CI with a force push: `git commit --amend --no-edit && git push --force`

bolshoytoster:
Oh it might be due to a change I made to `grow`. It used to be able to shrink allocations if `new_capacity < old_capacity`, but since the allocator API has separate `grow` and `shrink` functions (no `realloc`), I split it into two matching functions, and `grow` now panics if you give it a smaller capacity.

Should I revert the behaviour and just have a `if new_capacity < old_capacity { /*shrink*/ } else { /*grow*/ }`?

fereidani:
I'm a contributor like you, I think project members should comment on it.
But as your PR title is about custom allocator support, I would suggest to keep it as minimal as possible and direct to the point without no extra behavior changes.
Then check rust-lang/rust alloc behavior if smallvec does not behave the same, send another PR and fix that issue separately.

alejandro-vaz:
yeah probably just revert `grow` behavior with that if statement, it's what `realloc` does somewhat internally anyway

I have absolutely no idea what's happening with fuzzing. I haven't touched nor taken a look at it since I'm here, and I don't believe anyone here now has

I can't understand the traceback neither

I'll give it a look tomorrow

why doesn't equality allow for different allocators though??

bolshoytoster:
> why doesn't equality allow for different allocators though??

It's just `impl <T, U, A: Allocator, const N: usize, const M: usize> PartialEq<SmallVec<U, N, A>> for SmallVec<T, M, A>`, I wasn't sure if adding an `A2` would make it too much more complicated.

alejandro-vaz:
as long as it's correct why not do it??

alejandro-vaz:
did fuzzing solve itself or did I just miss something

bolshoytoster:
> did fuzzing solve itself or did I just miss something

The problem was the change in grow.

alejandro-vaz:
I've taken a look at the diff, looks good

I'll review it properly tomorrow.

src/rawsmallvec.rs
```diff
@@ -16,44 +17,48 @@ use {
...
-}
-
-impl<T, const N: usize> Default for RawSmallVec<T, N> {
-    #[inline]
-    fn default() -> Self {
-        Self::new()
-    }
+pub struct RawSmallVec<T, const N: usize, A> {
+    pub(crate) inner: RawSmallVecInner<T, N>,
+    pub(crate) alloc: A
```
alejandro-vaz, line 33: this looks too over engineered for our needs

why not keep `RawSmallVec` as it was but with the generic allocator parameter?? maybe we have to make the compiler like it with a phantomdata but that's better than having an inner union of the inner struct for the inside of the smallvec (too complex)

also, here `A` is unbounded, that's a problem

Cargo.toml
```diff
@@ -23,10 +23,12 @@ internals = []
 may_dangle = []
 rayon = ["std", "dep:rayon"]
 serde = ["dep:serde_core"]
+default = ["allocator-api2"]
```
alejandro-vaz: move `default` to the top for order

src/lib.rs
```diff
@@ -142,14 +151,13 @@ where R: core::ops::RangeBounds<usize> {
 }
 
 #[repr(C)]
-pub struct SmallVec<T, const N: usize> {
+pub struct SmallVec<T, const N: usize, A: Allocator = Global> {
     len: TaggedLen<T>,
-    raw: RawSmallVec<T, N>,
-    _marker: PhantomData<T>
```
alejandro-vaz, line 148: unexpected change, but yeah it should be removed

src/lib.rs
```diff
@@ -530,24 +539,23 @@ impl<I: Iterator, const N: usize> Drop for Splice<'_, I, N> {
...
     // # Safety
     //
     // `end` decides whether the data lives on the heap or not
     //
     // The members from begin..end are initialized
-    raw: RawSmallVec<T, N>,
+    raw: RawSmallVec<T, N, A>,
     begin: usize,
-    end: TaggedLen<T>,
-    _marker: PhantomData<T>
```
alejandro-vaz, line 542: ok

src/lib.rs
```diff
@@ -614,34 +622,22 @@ impl<T, const N: usize> DoubleEndedIterator for IntoIter<T, N> {
...
 
 impl<T, const N: usize> SmallVec<T, N> {
     #[inline]
     pub const fn new() -> SmallVec<T, N> {
-        Self {
-            len: TaggedLen::new(0, false),
-            raw: RawSmallVec::new(),
-            _marker: PhantomData
-        }
+        Self::new_in(Global)
```
alejandro-vaz, line 631: makes sense

alejandro-vaz (changes requested):

src/rawsmallvec.rs
bolshoytoster: We have to store the allocator, don't we?

alejandro-vaz: yeah, I just looked twice and the methods are not on the type, they are on the instance, you are right

it's kind of bullshit because most allocators like global are just a ZST but I guess it's just how the API works

I guess it's very ugly but it's the way

do you have any better idea??

alejandro-vaz: this change and #590 will certainly not get along well

alejandro-vaz: but I guess it'd just be a path change if I'm thinking about it right

alejandro-vaz:
everything looks right, we can extend this in the future I needed, this is an initial implementation

alejandro-vaz (approved):

src/rawsmallvec.rs
bolshoytoster: > it's kind of bullshit because most allocators like global are just a ZST but I guess it's just how the API works

Actually, one of the most popular custom allocators (566,924,100 all time downloads), bumpalo, implements `Allocator` for `&Bump`, which is not a ZST

> do you have any better idea??

If you want to keep `RawSmallVec` a union, we could just add `A` to both representations.

alejandro-vaz: oh I didn't knew bumpalo implemented allocator for a reference, that's interesting

adding `A` to both representations sounds much more confusing and tryhard than simply having an struct

alejandro-vaz:
is there any missing allocator support anywhere in the public API or are we basically done??

bolshoytoster:
There are the APIs I listed in the initial comment of the PR. Most of them are traits that can't be modified to include an `alloc: A` though, so it might only be possible to add counterparts for `from_buf*` and `from_raw_parts`.

We could also look back into `from_vec`.

It might also be possible to add it to `splice`, if you replace the `collect` with something else (since we can't implement it for `FromIterator`):
https://github.com/servo/rust-smallvec/blob/053f1a7b71b4ba1776ec6b47027f0ad7d92498a1/src/lib.rs#L520-L524

alejandro-vaz:
too complex for marginal feature additions

I have the impression that the whole allocator-api ecosystem in rust is half-baked and based on "so we add one generic and replace `alloc` for `A.allocate` whilst everything else that uses it indirectly breaks"

PR #593, opened 2026-09-08
Fix panic safety when reserving during splice

charliermarsh:
This PR fixes a double free when `splice` panics while reserving space:

```rust
use smallvec::SmallVec;

let mut v: SmallVec<Box<u8>, 4> = SmallVec::from_buf([Box::new(7)]);
drop(v.splice(0..0, std::iter::repeat_with(|| Box::new(1))));
```

While reserving, we temporarily increase the vector's length to include the elements we're keeping. If `reserve` panics, we skip resetting the length, so cleanup adds those elements a second time.

Instead, we can use `try_reserve` which allows us to reset the length before panicking.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -383,11 +383,13 @@ impl<T, const N: usize> Drain<'_, T, N> {
         let vec = unsafe { self.vec.as_mut() };
         let len = self.tail_start + self.tail_len;
 
-        // Test
+        // Include the tail when reserving so it survives a reallocation.
         let old_len = vec.len();
         unsafe { vec.set_len(len) }
-        vec.reserve(additional);
+        let result = vec.try_reserve(additional);
+        // Restore the prefix length before a reservation error can panic.
         unsafe { vec.set_len(old_len) };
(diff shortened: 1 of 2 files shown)


alejandro-vaz:
I'm withholding both of your PRs because of #590 and #592, I've seen them

thanks for contributing

charliermarsh:
No worries, thanks for the quick reply

alejandro-vaz (commented):
do these tests fail with the old implementation??

charliermarsh:
I believe `splice_reserve_panic` does (`splice_spill_preserves_tail` passes on both).

alejandro-vaz:
does `splice_spill_preserves_tail` fail under miri with the old implementation because it detects a double free??

charliermarsh:
It doesn't as-is, but it can be made to fail under Miri (before the library change) by boxing it.

alejandro-vaz (approved):

alejandro-vaz:
thanks for contributing

PR #594, opened 2026-09-08
Fix element ownership tracking with may_dangle

charliermarsh:
With `may_dangle` enabled, we currently allow references read by an element's destructor to expire before the vector is dropped, causing a use-after-free.

Track ownership of `A::Item` with `PhantomData` so drop checking rejects this, including spilled vectors with zero inline capacity. Plain borrowed references without destructors retain the intended `may_dangle` behavior.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -113,8 +113,6 @@ extern crate std;
 #[cfg(test)]
 mod tests;
 
-#[cfg(feature = "serde")]
-use core::marker::PhantomData;
 #[cfg(feature = "drain_keep_rest")]
 use core::mem::ManuallyDrop;
 #[cfg(feature = "malloc_size_of")]
@@ -140,6 +138,7 @@ use {
         hash::{Hash, Hasher},
         hint::unreachable_unchecked,
         iter::{repeat, FromIterator, FusedIterator, IntoIterator},
+        marker::PhantomData,
         mem::{self, MaybeUninit},
         ops::{self, Range, RangeBounds},
         ptr::{self, NonNull},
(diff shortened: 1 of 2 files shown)


alejandro-vaz:
this and #596 won't make it to v1.16.1, but we can make a second patch later

charliermarsh:
Cool, thanks for the heads up

alejandro-vaz:
yeah I was looking at this the other day whilst doing some refactoring and it looked weird to me

does `may_dangle` even make sense there?? is it correct??

I have an intuition it is not, though I can't tell for sure

we are reading the references in order to run the destructors, so the reference must be valid and the element not dropped, which means that dropping the element before the smallvec drops leads to us reading a dangling reference (even if later we aren't double freeing)

I haven't personally used `may_dangle` until now so I'm somewhat confused

alejandro-vaz (commented):
that phantomdata was exactly what @bolshoytoster removed on v2 if I remember correctly

bolshoytoster:
I did not know about this behaviour. v2 should be fine though, since `TaggedLen` still stores a `PhantomData<T>`.

alejandro-vaz (commented):
> yeah I was looking at this the other day whilst doing some refactoring and it looked weird to me
> 
> does `may_dangle` even make sense there?? is it correct??
> 
> I have an intuition it is not, though I can't tell for sure
> 
> we are reading the references in order to run the destructors, so the reference must be valid and the element not dropped, which means that dropping the element before the smallvec drops leads to us reading a dangling reference (even if later we aren't double freeing)
> 
> I haven't personally used `may_dangle` until now so I'm somewhat confused

alejandro-vaz (approved):
just researched how `may_dangle` actually works

it is correct now

thanks for contributing

PR #596, opened 2026-09-08
Backport v2’s retain implementation to v1

charliermarsh:
## Summary

This PR applies the existing v2 compaction approach to v1's `retain`: leave the initial retained prefix in place, then move each survivor into the next hole instead of swapping it with a discarded element. Removed elements are dropped during the scan, and a guard restores the untouched tail if the predicate or a destructor panics.

I benchmarked this on Linux (x86), and retaining half the input took 40-60% less time for small integer vectors, 16-21% less time for 4,096 integers, and 24-38% less for 4,096 larger elements:

| Workload | Element size | Before | After | Paired time change |
|---|---:|---:|---:|---:|
| Keep half, 16 elements | 8 B | 39.33 ns | 22.10 ns | −45.5% to −42.0% |
| Keep half, 17 elements | 8 B | 46.95 ns | 20.97 ns | −59.9% to −54.1% |
| Keep half, 4,096 elements | 8 B | 17.205 µs | 14.247 µs | −21.0% to −15.9% |
| Keep half, 4,096 elements | 128 B | 31.931 µs | 21.497 µs | −37.8% to −23.6% |
| Keep all, 16 elements | 8 B | 39.79 ns | 12.46 ns | −69.8% to −67.6% |
| Keep all, 17 elements | 8 B | 38.71 ns | 9.93 ns | −80.3% to −73.0% |
| Keep all, 4,096 elements | 128 B | 19.722 µs | 14.650 µs | −29.7% to −7.3% |
| Keep none, 16 elements | 8 B | 40.60 ns | 17.24 ns | −57.6% to −55.8% |
| Keep none, 17 elements | 8 B | 40.94 ns | 18.55 ns | −55.1% to −54.4% |
| Keep none, 4,096 elements | 128 B | 18.427 µs | 17.594 µs | −7.0% to +13.1% |
| Unchanged `Vec` control: keep half, 4,096 elements | 128 B | 21.275 µs | 21.486 µs | −9.6% to +10.8% |

(I included some native benchmarks but removed in a follow-up commit; you can see them [in the PR history](https://github.com/astral-sh/rust-smallvec/blob/88c6bfabd610f34f4f1f9d530f58fb8a74b9a807/benches/retain.rs).)


tests/retain.rs
alejandro-vaz, line 1: do not create a `tests/retain.rs` file for that, include it on `src/tests.rs`

alejandro-vaz (changes requested):

src/tests.rs
```diff
@@ -786,6 +786,106 @@ fn test_retain() {
     assert_eq!(Rc::strong_count(&one), 1);
 }
 
+mod retain {
```
alejandro-vaz: can this be done without a module please??

just bare on `src/tests.rs` as a set of tests inside the file at the root

alejandro-vaz (changes requested):

src/tests.rs
charliermarsh: Sure no worries -- I figured this was okay since we already have `mod insert_many_panic` here and this gives the structs their own namespace. Will change.

alejandro-vaz: hahahah I did not know about `mod insert_many_panic`

v1 is all over the place, I prefer not looking at it twice, it barely makes sense

thanks for changing it though

alejandro-vaz (approved):

alejandro-vaz:
can you paste the table with performance changes??

just to have it explicit on the PR

charliermarsh:
Done!

alejandro-vaz:
thanks @charliermarsh

PR #610, opened 2026-09-14
Add testing and benchmarking steps with `--no-default-features`

### LINKED ISSUE #605: add `--no-default-features` steps to CI
on `.github/workflows/rust.yml`, add two steps that only start on nightly that run testing and benchmarking with `--no-default-features`

this has become needed due to the recent addition of `allocator-api2` as a default feature

pedrodesu:
Fixes https://github.com/servo/rust-smallvec/issues/605.


pedrodesu:
The code at the previous point imported `Allocator` and `Global` incorrectly from `alloc` (instead of `alloc::alloc`), thus it failed. Additionally, we must force the `allocator_api` feature if we're not using the default `allocator-api2` feature.

I also changed the extern of `std`, it's redundant since it's under the feature of `std`. I believe it's better to just invert the logic and only toggle `#![no_std]` if the `std` feature isn't present.

My PR reflects these changes.

One thing to bear in mind: Our current code forces the usage of an allocator API, whether nightly `allocator_api` or `allocator-api2`'s. This means that if a user on stable rust disables `allocator-api2`, our library would never compile as-is (indeed, if you hypothetically ran `Cargo test no default features` under stable, surely the action would fail).
Nothing needs to be changed if that's deliberate, but if not, we should probably add more conditional code to fallback to the API of no custom allocators if we don't detect either `nightly` or `allocator-api2`. This, of course, has the price of more code.

alejandro-vaz (approved):

alejandro-vaz:
thanks, looks good.

PR #611, opened 2026-09-14
Remove deprecated usage in `benches/bench.rs`

pedrodesu:
As https://github.com/servo/rust-smallvec/pull/589 removes the usage of the deprecated `smallvec` macro, I took the liberty to write this small PR to also replace it in the benchmark and rid it of its `#![allow(deprecated)]`.

This refactor revealed an underlying problem: `vec!` is not deprecated, and allows the notation `vec![T; N]`. `smallvec!` IS deprecated, and there's no equivalent in this regard;
- `smallvec::from_elem(val, n)`, what I use (internally) in this PR, is hidden from documentation and isn't meant to be consumed by the public. Its doc comment literally states "It is recommended to use the macro instead of using thís function." (which isn't the case anymore as we know, since the macro got deprecated). The `std` does this as well, but the difference is that the `std` gives the `vec!` macro as the recommendation and didn't deprecate it.
- `SmallVec::from`, the suggested counterpart, obviously doesn't fit because a) In `SmallVec::from([val; n]);`, `[val; n]` won't compile if `n` isn't known at compile-time, and b) `SmallVec::from(vec![val; n]);` is obviously self-defeating and is fundamentally a different case, as we're transforming a `Vec` into a `SmallVec`, not constructing a `SmallVec` from a value and a length.

To summarise, this gives us the question: What is a consumer supposed to do if they want to make a `SmallVec` out of `n` elements of `val`, not being `n` a constant, and without using a deprecated macro?

I believe we might want to answer this question. Perhaps de-deprecating the macro, giving a specific function or something else.


alejandro-vaz (approved):

alejandro-vaz:
makes a lot of sense, didn't realize we still had use of the macro

thanks for contributing

alejandro-vaz:
by the way, feel free at any time to open PRs without any issue being required

PR #619, opened 2026-09-14
Refactor the benchmark suite

TDecking:
The `make_benches!` macro will now create only one function containing the benchmarks and will put all bechmarks into a benchmark group. This will remove a bit of bloat while making the report produced by criterion easier to read.


alejandro-vaz:
it makes benchmarking way nicer

thanks for your contribution

alejandro-vaz (approved):

Issue #627, opened 2026-09-19
`lib.rs` iterator modularization


alejandro-vaz:
`lib.rs` not only contains the implementation of `SmallVec`, but all its iterators as well: `Drain`, `ExtractIf`, `IntoIter`, `Splice` (except the parallel iterator)

the goal is to move each struct with its implementations into a file under `src/iterators/` which is already created

expect conflicts regarding visibility

this comes originally from #479 but was cancelled due to its complexity back then. it probably is as complex now but we have to do it anyway

Kxrma47:
I can take Drain and ExtractIf into src/iterators/, including visibility fixes and tests. Would it be useful to split IntoIter and Splice with another contributor and coordinate one joint PR, or would you prefer separate PRs? I will first check the existing iterator tests and keep behavior unchanged. Happy to follow your preferred split before starting.

alejandro-vaz:
separate PRs are easier to merge, though either way is fine

I already tried before and what I can remember is that visibility becomes a nightmare, though codebase might have changed since then. so yeah go start now and see what you come across with

alejandro-vaz:
I've assigned the issue to you

alejandro-vaz:
brother aren't we missing intoiter and splice as well??

are they more complex or something??

Kxrma47:
Yes, IntoIter and Splice are still in scope. I started with Drain and ExtractIf as separate PRs to keep each change reviewable, as you suggested. I will prepare separate PRs for IntoIter and Splice too, preserving the public API and existing behavior. Splice uses Drain, so I will account for that dependency and coordinate the shared imports when the pending PRs merge. I have not established that either is inherently too complex; they were simply the next pieces of the issue.

alejandro-vaz:
... are you still working on the `IntoIter` and `Splice` PRs??

Kxrma47:
Completed the iterator split: #628 moves Drain, #629 moves ExtractIf, #666 moves IntoIter, and #667 moves Splice.

PR #628, opened 2026-09-19
Move Drain iterator into its own module

Kxrma47:
Moves Drain and its iterator, Drop, and Debug implementations from lib.rs to src/iterators/drain.rs. Keeps the public smallvec::Drain export; crate-visible fields and methods preserve the existing SmallVec::drain and Splice integration without changing behavior.

Part of #627. ExtractIf will follow separately.

Validation: cargo test; cargo test --features 'std rayon serde borsh bytes arbitrary'; cargo clippy --all-targets -- -D warnings; cargo doc --no-deps. Nightly rustfmt is exercised by CI.


src/iterators/mod.rs
```diff
@@ -1,2 +1,5 @@
+mod drain;
+pub use drain::Drain;
```
alejandro-vaz: honestly I think it'd be simpler to have `pub mod drain;` and then on `lib.rs` have a `pub use iterators::drain::Drain;`

src/iterators/drain.rs
```diff
@@ -0,0 +1,204 @@
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
```
alejandro-vaz: no need to include the license

alejandro-vaz (approved):

alejandro-vaz (changes requested):
changes above

alejandro-vaz (approved):

src/iterators/mod.rs
Kxrma47: Updated as suggested: `iterators/mod.rs` now declares `pub mod drain;`, and `lib.rs` re-exports `iterators::drain::Drain`. I applied the same module pattern to the separate ExtractIf PR (#629).

src/iterators/drain.rs
Kxrma47: Removed the duplicated license header from `drain.rs` (and from `extract_if.rs` in #629). Thanks for the review.

PR #641, opened 2026-09-20
ci: remove the SmallBot workflow

### LINKED ISSUE #638: delete `.github/workflows/smallbot.yml`
`smallbot` was a fun experiment but no one really uses it and consumes CI

why would you use it if you can talk to a human being, it makes zero sense, of course people just ask to be assigned

bye smallbot

Denimworld12:
Closes #638.

`smallbot.yml` is self-contained — the only occurrences of "smallbot" in the repository are inside that file, so nothing else needs updating. Deleting it is the whole change.

For the record, it handled two commands on issue comments:

- `smallbot! assign` — assign the commenter, if the issue had no assignee
- `smallbot! unassign` — remove the commenter

It also triggered on `issue_comment: [created]`, so a job was queued for every comment on every issue and exited immediately unless the body started with `smallbot! `.


alejandro-vaz (approved):

alejandro-vaz:
looks good, thanks for contributing

Issue #632, opened 2026-09-19
fix allocator API `into_boxed_slice` and `into_vec`


alejandro-vaz:
> ```rust
>     #[inline]
>     pub fn into_boxed_slice(self) -> Box<[T]> {
>         self.into_vec().into_boxed_slice()
>     }
> ```
> 
> and also
> 
> ```rust
> pub fn into_vec(self) -> Vec<T> {
> ```
> 
> Both are UBs.

Whoops, I thought I left all the stdlib-interacting functions only implemented for `SmallVec<_, Global>` (`from_vec` is correct).

_Originally posted by @bolshoytoster in https://github.com/servo/rust-smallvec/issues/630#issuecomment-5744461318_

alejandro-vaz:
and is there any special reason there's no `From<SmallVec<T, N, A>> for Vec<T, A>` and so on?? to me it looks like missing APIs

bolshoytoster:
> and is there any special reason there's no `From<SmallVec<T, N, A>> for Vec<T, A>` and so on?? to me it looks like missing APIs

That won't work when using `allocator-api2`, they'd be using different `Allocator` traits.

alejandro-vaz:
no, `allocator_api2::vec::Vec` solves it

but yeah it's a different `Vec`, not an interoperability, and I don't think it should be

alejandro-vaz:
we should maybe include both or something like that, I'm not fully sure

we have made `SmallVec` to be allocator-first but the rest of rust hasn't quite caught up

PR #649, opened 2026-09-21
Add From<SmallVec> implementations for Vec and Box<[T]>

### LINKED ISSUE #633: add missing `From` conversions to `SmallVec`
those which have already `into_*` methods, but for some reason we don't have them as convert implementations

e.g.:
1. to `Vec<T, A>`
2. to `Box<[T], A>`

etc etc etc

that have methods but not the `From` trait conversion

shuaixr:
Closes: #633 
Implemented the `From` trait for converting `SmallVec` into `Vec` and `Box<[T]>` using the existing `into` methods, and added tests for both.

feedback and suggestions are welcome!


alejandro-vaz (commented):
this PR is blocked by https://github.com/servo/rust-smallvec/issues/632

we're still not sure of how to fix it but the logical fix would be to have two box and two vec conversions, one for the normal alloc and the other for the allocator-api supplied box/vec

shuaixr:
Oh yes, I overlooked that.
I can update the implementation once #632 is resolved.

alejandro-vaz:
I'm making the PR as a draft since it is blocked

to have it mentally checked in

alejandro-vaz:
okay all fuzzing and new CI and allocator features have landed

we finally agreed on something simpler: all the conversions that use `Vec` and `Box` will use `crate::Vec` and `crate::Box` that come from our custom allocator backend

that means, by default and with the feature `allocator-api` it uses `alloc::{vec::Vec, boxed::Box}`, and with the feature `allocator-api2` it uses `allocator_api2::{vec::Vec, boxed::Box}`

can you update with this??

shuaixr:
Yeah, I’ve updated it. Please take a look

tests/main.rs
```diff
@@ -852,6 +852,28 @@ fn from_vec() {
...
+fn box_from_small_vec() {
+    let small_vec = SmallVec::<u8, 2>::from_iter(0..2);
+    let boxed_slice = Box::<[u8]>::from(small_vec);
+    assert_eq!(&*boxed_slice, &[0, 1]);
+
+    let small_vec = SmallVec::<u8, 2>::from_iter(0..3);
+    let boxed_slice = Box::<[u8]>::from(small_vec);
+    assert_eq!(&*boxed_slice, &[0, 1, 2]);
+}
+
```
alejandro-vaz: why are these tests necessary??

alejandro-vaz: don't we have similar ones??

shuaixr: yes they a bit redundant. I’ll remove it

alejandro-vaz: I'm planning to deprecate the `into_*` methods soon so naturally the other tests will migrate to use `Into::into` anyway

alejandro-vaz (approved):

alejandro-vaz:
thanks for contributing @shuaixr

PR #658, opened 2026-09-22
Re-implement fuzzing via libfuzzer

### LINKED ISSUE #657: remake fuzzing from scratch
our current fuzzing is obscure and no one really knows how it works

it is under `fuzz/`

it was back from when the repo didn't even use GitHub actions, it used travis-CI, for perspective

and I almost broke it with #656 

simple: take it down and build new fuzzing, probably in the form of some sort of special test

implementation details are left for you to decide

pedrodesu:
Closes https://github.com/servo/rust-smallvec/issues/657.

This might need some additional work; I'm treating this PR more like a draft.
Additionally, I've already ran this fuzzer myself and it has already shown quickly that we have some violations occuring with our current code on `v2` (this assumes that the current fuzzing implementation is proper, which I believe it is), namely on `try_grow`'s `assert!(new_capacity >= len);` and on the fuzzing test's assertion that `assert_eq!(small_vec.spilled(), small_vec.capacity() > N);`. From what I saw, it's not really failing anywhere else. This also means, of course, that the fuzzing action should naturally fail at this point (I've also changed the fuzzing CI, I think it's working).

Small disclaimer: I've done this work through the small hours of the morning lol


pedrodesu:
Ah, this refactor not only made the implementation much more robust and simple, both in terms of code and running (via `libFuzz` and `cargo-fuzz`), but it also improves on the further maintainability of the code and on the actual fuzzing procedure, namely the biggest win being that we now not only check if executing operations panic/crash, but we also test differences between `SmallVec` and std's `Vec`.

pedrodesu:
I also don't think it's the best idea to do either fuzzing or benchmarking on PRs. It takes a considerable amount of time (namely for the benchmarking part), the fuzzing is flaky/inconsistent, especially for a 30 secs run, and the benchmarking is probably noisy and unreliable given that it's running on GitHub Actions runners, where it probably shouldn't run. I feel like we should run an extensive, longer, proper fuzzing under a cron schedule and benchmark manually via comments or using dedicated runners instead. But this is an off-topic thing I noticed during the nightly action run and, if relevant, should probably have an issue of its own.

fuzz/Cargo.toml
```diff
@@ -1,26 +1,23 @@
 [package]
 name = "smallvec-fuzz"
 version = "0.1.0"
-authors = ["Dawid Ciężarkiewicz <dpc@dpc.pw>"]
-edition = "2021"
+authors = ["The Servo Project Developers"]
 publish = false
+edition = "2024"
```
alejandro-vaz, line 6: + rustc-version = 1.86 for MSRV

fuzz/Cargo.toml
```diff
@@ -1,26 +1,23 @@
 [package]
 name = "smallvec-fuzz"
 version = "0.1.0"
```
alejandro-vaz: let's set it to 0.0.0 instead

fuzz/.gitignore
```diff
@@ -0,0 +1,4 @@
+target
+corpus
+artifacts
+coverage
```
alejandro-vaz, line 4: why not have it instead of `./.gitignore`?? it'd be simpler to maintain than another file

fuzz/in/stub
alejandro-vaz, line 1: what was even the purpose of this file?? wtf

.github/workflows/rust.yml
alejandro-vaz, line 1: this should be ported to #644 as well

fuzz/Cargo.toml
```diff
@@ -1,26 +1,23 @@
...
-[features]
-afl_fuzz = ["afl"]
-honggfuzz_fuzz = ["honggfuzz"]
-
-
 [dependencies]
-honggfuzz = { version = "0.5.47", optional = true }
-afl = { version = "0.4", optional = true }
-smallvec = { path = ".." }
+arbitrary = { version = "1.4.2", features = ["derive"] }
```
alejandro-vaz: for consistency with our main vendor file, let's set version to major.minor here as well 

"1.4"

fuzz/README.md
alejandro-vaz, line 1: I think this README is completely unnecessary if CI already runs fuzzing

and the other info are just implementation details we don't have to disclose here

fuzz/fuzz_targets/smallvec_ops.rs
```diff
@@ -1,237 +1,336 @@
...
+    Pop,
+    Grow(Bounded<usize>),
+    Reserve(Bounded<usize>),
+    ReserveExact(Bounded<usize>),
+    ShrinkToFit,
+    Truncate,
+    SwapRemove,
+    Clear,
+    Remove,
+    Insert { val: usize },
```
alejandro-vaz: this should be a tuple struct probably

fuzz/fuzz_targets/smallvec_ops.rs
```diff
@@ -1,237 +1,336 @@
...
+    Drain,
+    Splice { items: Bounded<Vec<usize>> },
```
alejandro-vaz: same here

fuzz/fuzz_targets/main.rs
alejandro-vaz, line 1: could we:
- rename this file to something like `fuzzing.rs`
- add it as a normal bin for `./Cargo.toml` instead under `./fuzz/fuzzing.rs.rs`
- remove the additional package
- remove the unnecessary `fuzz_targets/` dir

??

that would make it much simpler

alejandro-vaz (changes requested):

alejandro-vaz:
also, there is a weird interaction with #656 which is not nice at all, the API there is a bit fucked up currently

fuzz/Cargo.toml
pedrodesu: do we wanna duplicate this for both `Cargo.toml`s? doesn't seem like a great idea. additionally, fuzzing requires nightly. the fuzz package should definitely not adhere to the root's MSRV. we should either omit or pin a high rustc-version. I feel like omitting is best.

fuzz/.gitignore
pedrodesu: I didn't understand. Did you mean just removing this file and adding these to the root's .gitignore? We can do this, I preferred to keep this .gitignore initially because it's the way that `cargo fuzz init` does and ultimately reinforces the idea that they're 2 different separate packages and that they have a division of concerns (as in, fuzz's ignore files are exclusive to this package and go here). But I can merge them into the root .gitignore if you think it's a good idea.

fuzz/in/stub
pedrodesu: I think it might've been a leftover residual file of a test or a previous fuzz iteration. Idk

.github/workflows/rust.yml
pedrodesu: I think it'd be a good idea to keep the CI as is in this PR and to solve https://github.com/servo/rust-smallvec/pull/644 separately (and ultimately its closing will overwrite our current CI, including the fuzzing part). This way we don't have to keep syncing their changes or going back and forth. This CI should be good enough (in regards to fuzzing) for the time being.

fuzz/README.md
pedrodesu: I can shrink the README or even remove it, though I do think it is beneficial to at least explain that we're using cargo-fuzz and libFuzzer and on how to run the fuzzer (via cargo fuzz run). I feel like this is bare minimum, else anyone who stumbles upon the fuzzing has to check the code and guess what we're using to do the fuzzing, realize that we're using cargo fuzz and check how to run it if they aren't familiar with cargo fuzz yet.

alejandro-vaz: yeah but fuzzing is not meant to be run manually

no one is going to wait for four hours on their machine doing some fuzzing

it's only for CI

brother I maintain this crate and haven't triggered fuzzing manually once

I'd delete it

fuzz/.gitignore
alejandro-vaz: yeah I mean removing it and have it in the root `.gitignore`

part of this https://github.com/servo/rust-smallvec/pull/658#discussion_r4070490120 as well

fuzz/Cargo.toml
alejandro-vaz: thinking it better, if we are thinking of having fuzzing as an special binary of the normal `Cargo.toml`, it just means merging

fuzz/fuzz_targets/main.rs
pedrodesu: 1) We can, but I wouldn't recommend. It's redundant. We'd be running `cargo fuzz run fuzzing`. Something like the current `smallvec_ops.rs` gives us much more to work with.

3) No. Under `cargo-fuzz`, fuzzing must be done under a separate package. Indeed it makes sense, as it means gatekeeping additional unnecessary dependencies (namely `libfuzzer-sys`) to the root `Cargo.toml`, and because each harness file compiles to a separate, individual binary linked against `libFuzzer` (this is why we have `#![no_main]`). In other words, fuzzing is in itself a completely different operation from the main library, and thus should have its own space. It mustn't be "integrated" on the main crate, and its internal functioning disallows that.

4) We can (by changing `[[bin]]`). However, because it's a different, standard rust crate (as explained in 3)), I'd advise to keep some sort of folder (ideally `fuzz_targets/`, see the footnote). It's 1) standard practice for rust crates and 2) a good idea because once you start fuzzing the fuzz crate will get polluted with residual folders.

2) We can (by changing the root's `[[bin]]`). However, we definitely shouldn't. If we do that, then our main root will have `#![no_main]` binaries that expect `main` and other symbols to be linked via `libFuzzer`. This means that actions like `cargo check`, testing and testing under miri could fail if not under `nightly` because for the fuzzing to compile we need nightly; Additionally, once again, you'll bloat the main crate with more unnecessary libraries (namely `libfuzzer-sys`).

Additionally if when tackling #644 we intend to, for instance, use an automated fuzzing service, we will need to use the canonical Rust fuzzing layout (which means having the separate crate under `fuzz/` and the structure `fuzz_targets/*.rs`). That seems to be the standard structure of fuzzing in Rust and of `cargo-fuzz`, so I would recommend adhering to it.

alejandro-vaz: okay then leave it as-is

fuzz/Cargo.toml
alejandro-vaz: oh I still think we should add `rust-version.workspace = true` here just in case

and have the version configured in the workspace

alejandro-vaz (changes requested):

fuzz/README.md
pedrodesu: Fair enough

fuzz/.gitignore
pedrodesu: Since we will keep the fuzzing package separate, we have an even stronger case to leave the `.gitignore` inside of it. I can merge it to the root `.gitignore`, but I'd discourage it. Let me know what you think.

alejandro-vaz: doesn't really matter, leave it as-is if you want

fuzz/Cargo.toml
pedrodesu: This indeed solves the question of duplicating the `rustc-version` between two packages, but it doesn't address the other problems. That would imply that fuzzing works under `rustc 1.86`, which it *does not*. Fuzzing only works on nightly. Because of this, this could also theoretically lead to toolchain or dependency resolution errors when fuzzing.

alejandro-vaz: oh I didn't know fuzzing only worked on nightly, okay

fuzz/fuzz_targets/main.rs
alejandro-vaz: so this file didn't get renamed to `fuzzing.rs`??

alejandro-vaz (approved):

alejandro-vaz (changes requested):
above

fuzz/Cargo.toml
pedrodesu: Yeah, in fact that's why I changed the CI to only run fuzzing if on nightly.

As `cargo-fuzz` states in their README:
> Note: libFuzzer needs LLVM sanitizer support, so this only works on x86-64 and Aarch64, and only on Unix-like operating systems (not Windows). This also needs a nightly compiler since it uses some unstable command-line flags. You'll also need a C++ compiler with C++11 support.

fuzz/fuzz_targets/main.rs
alejandro-vaz: or idk, something like `main.rs`, whatever goes

fuzz/Cargo.toml
alejandro-vaz: those are a lot of preconditions

fuzz/fuzz_targets/main.rs
pedrodesu: > We can, but I wouldn't recommend. It's redundant. We'd be running `cargo fuzz run fuzzing`. Something like the current `smallvec_ops.rs` gives us much more to work with.

I suppose `main.rs` could be an alternative, but "main" generally implies a standard application entry point. A fuzz target is a specialized harness. Additionally, if in the future we add more targets, to for instance support fuzzing with custom flags or behavior or something of the sort, `main.rs` could be unclear. It's the same rationale as with testing: I don't think a `main.rs` test file is a great practice, but it's a possibility, and indeed our library has such a file. In the same manner, we could either keep `smallvec_ops.rs`, or go with `main.rs` for consistency with the testing.

alejandro-vaz: I'd go with `main.rs` and spend zero more thoughts about it

alejandro-vaz:
how do we now interpret the weird fuzzing error in https://github.com/servo/rust-smallvec/actions/runs/35758316781/job/106849613954??

I'm not sure what debug has to do there

pedrodesu:
> how do we now interpret the weird fuzzing error in https://github.com/servo/rust-smallvec/actions/runs/35758316781/job/106849613954??
> 
> I'm not sure what debug has to do there

That's just the `Debug` output of the input byte sequence generated by `libFuzzer` when the target crashed.
What actually failed is written earlier:
```sh
thread '<unnamed>' (10010) panicked at src/lib.rs:892:9:
assertion failed: new_capacity >= len
```
The fuzzer generated a sequence of operations that forced `SmallVec` into a state where `new_capacity` dropped below `len`, triggering an assertion failure.

You should find that failure locally by fuzzing in your machine and accessing the crash file like the output of the action suggests. We could also change the CI action so that if fuzzing fails it gives us the crash either via the action's artifacts or by printing it directly in the fail. That's probably a good idea. As of now our fuzzing action tells us if something fucks up, but we can't actually access the crash to reproduce it ourselves in our machines. But as I said at the start of the issue, the fuzzing seems to fail always on the same two situations, so you can probably get the fail and access the crash yourself by simply running the fuzzing in your machine as of now.

alejandro-vaz:
how is that even possible

has any invariant been broken??

pedrodesu:
Apparently, yes. The invariant that a `len`, or a to-be-formed, would-be `len`, `<= cap`, concretely. But I haven't looked too deep into the panic yet and I'm not on the computer right now. But the way that I wrote the fuzzer is to, among other things, verify invariants. This is the case assuming that, of course, this isn't a false positive from the fuzzer, which it doesn't seem like it is and shouldn't be if the fuzzing is properly done, which it seems like it is.

I will look into this at my earliest convenience. If we can indeed prove that it's a code problem, then this fuzzer is already finding existing problems that we hadn't caught and it should be a good idea to merge. Independently of merging then directly or not, the end goal is to, of course, let the fuzzer run its course for a lot of time without finding any crash or panic.

alejandro-vaz:
great I guess...

I already doubt many parts of our implementations and I'm planning to drop soon an issue group for 10-15 issues regarding code revision, this seems to confirm it

I wonder what would the fuzzer find if ported to v1, and I think we should do it if possible

go investigate the crash whenever you have time if you want, if you don't find time I'll take it no worries

fuzz/fuzz_targets/main.rs
```diff
@@ -0,0 +1,332 @@
...
+    Clear,
+    Remove,
+    Insert(usize),
+    Drain,
+    Splice(Bounded<Vec<usize>>),
+    RetainEven,
+    Dedup,
+    ExtendFromSlice(Bounded<Vec<usize>>),
+    ExtendFromWithin,
+    Resize { new_len: Bounded<usize>, val: usize }
```
alejandro-vaz: hmmm maybe this should be a tuple structure as well to be honest

alejandro-vaz (changes requested):

fuzz/fuzz_targets/main.rs
```diff
@@ -0,0 +1,332 @@
...
+    assert_eq!(
+        small_vec.as_slice(),
+        std_vec.as_slice(),
+        "`as_slice()` mismatch"
+    );
+    assert_eq!(
+        small_vec.as_mut_slice(),
+        std_vec.as_mut_slice(),
+        "`as_mut_slice()` mismatch"
+    );
```
alejandro-vaz: either is unnecessary

if the immutable references are equal, the mutable references will as well

I'd only keep the immutable one honestly

fuzz/fuzz_targets/main.rs
```diff
@@ -0,0 +1,332 @@
...
+        small_vec.capacity() > N,
+        "`spilled()` doesn't equal to `capacity() > N`"
+    );
+
+    // Indexing and bounds invariants
+    for i in 0..small_vec.len() {
+        assert_eq!(
+            small_vec[i], std_vec[i],
+            "`small_vec[{i}]` doesn't match `std_vec[{i}]`"
+        );
```
alejandro-vaz: this is unnecessary

if both immutable slices are equal as checked above, all elements are equal

fuzz/fuzz_targets/main.rs
```diff
@@ -0,0 +1,332 @@
...
+        assert_eq!(
+            small_vec.get(i),
+            std_vec.get(i),
+            "`small_vec.get({i})` doesn't match `std_vec.get({i})`"
+        );
```
alejandro-vaz: same here

fuzz/fuzz_targets/main.rs
```diff
@@ -0,0 +1,332 @@
...
+    }
+    assert_eq!(
+        small_vec.get(small_vec.len()),
+        None,
+        "out-of-bounds `get()` did not return `None`"
+    );
```
alejandro-vaz: this by definition will always pass

alejandro-vaz (changes requested):

alejandro-vaz:
also, CI is complaining that we renamed `smallvec_ops.rs` to `main.rs` without doing the proper rename elsewhere

pedrodesu:
Yes, it's in the `rust.yml`. Fixing that.

fuzz/fuzz_targets/main.rs
pedrodesu: The thing is - they will both be equal *in theory*. In theory, all of the assertions done in fuzzing should pass, because all of them attempt to guarantee that each invariant isn't violated. `as_mut_slice` and `as_slice`, albeit very similar in their internal implementation, are different functions. What happens if in the future we change or try to change one's implementation but don't change the other? What if one of the invariants is ever broken and we don't catch it? What do we lose by asserting that both match? We do lose a bit of certainty by asserting that only one matches. Does this make sense? My point is, we probably shouldn't "stop doing an assertion" because we can imply that it passes from other assertion or because it's "simply logical". Apparently, it's also "simply logical" that `cap` would always be greater or equal to `length`, but the previous failing fuzzing output shows that that's apparently not the case.

tl;dr I'd insist we keep `as_mut_slice`, indexing, `get` and bounds check. By definition they all pass, until they don't and we have edge cases with UB or crashing. We lose nothing by keeping these assertions. Still, if you'd still prefer to remove them, I can change the code in that manner.

fuzz/fuzz_targets/main.rs
pedrodesu: Please reply to the comment above.

fuzz/fuzz_targets/main.rs
pedrodesu: Please reply to the comment above.

fuzz/fuzz_targets/main.rs
pedrodesu: Please reply to the comment above.

pedrodesu:
Btw, I think I figured out the failure that the last CI was giving. We have `assert!(new_capacity >= len);` on a public function (`try_grow`). `new_capacity` is an argument. My guess is that the fuzzer most definitely called the function `try_grow` via `grow` via `Op::Grow` with a `new_capacity` smaller than `len`, which is why this panic happened. What stops anyone from replicating this? My guess is that we shouldn't `panic`/`assert`, but rather do nothing or return an `Err`, as `try_grow` is already supposed to (as it returns a `Result`).
Additionally, since this function is public, why do we mark it as `#[cold]`?

fuzz/fuzz_targets/main.rs
alejandro-vaz: in the case they stop doing what they are meant to do, tests will break before fuzzing matters

honestly I get the point, but they are so low-level invariants that at this point we may as well check that taggedlen isn't lying about the heap invariant. it's like, if this fails then everything else will fail before so it really makes no sense to explicitly check it

pedrodesu:
Indeed, I patched the function up quickly by returning a dummy `Ok(())` if `new_capacity < len` (Of course we should probably return an `Err` instead or something similar if we do indeed stop the `panic`) and the fuzzing *doesn't seem* to complain anymore about this case. At least by running locally, I observe that it's only complaining about the other assertion:
```
assertion `left == right` failed: `spilled()` doesn't equal to `capacity() > N`
  left: true
 right: false
 ```

**Comment by alejandro-vaz** [comment]
I honestly can't reply to that, who knows, it was already there when I came here

I'm going to open full code revision soon because it doesn't make much sense to me either

I'm not exactly sure what it should do but probably not return another error. either don't do anything or make it a precondition or something like that

**Comment by alejandro-vaz** [comment]
> Indeed, I patched the function up quickly by returning a dummy `Ok(())` if `new_capacity < len` (Of course we should probably return an `Err` instead or something similar if we do indeed stop the `panic`) and the fuzzing _doesn't seem_ to complain anymore about this case. At least by running locally, I observe that it's only complaining about the other assertion:
> 
> ```
> assertion `left == right` failed: `spilled()` doesn't equal to `capacity() > N`
>   left: true
>  right: false
> ```

okay... why...??

**Comment by pedrodesu** [comment]
I don't know yet. From what I ran, the fuzzing complained about these two specific cases. the `try_grow` one seems to be discovered and ready to be solved, and it's honestly less worrying because it's a simple `assert` being fired that we wrote ourselves in the codebase. This one, however, is apparently triggered by an invariant violation via an assertion in the fuzzer itself. I don't know, I'll try to replicate it and understand why this is happening. In the meantime, I'll push the agreed upon changes (Note: I didn't touch the `try_grow` function in this change. It's quite possible that the fuzzer will complain about it again as it did before.)

**Comment by alejandro-vaz** [comment]
I agree but in order to make sure CI is happy let's simply comment out those assertions, we'll re-enable them as we fix root causes

**Comment by alejandro-vaz** [comment]
and conflicts with the workflow have appeared, rebase is needed

**Comment by pedrodesu** [comment]
> I agree but in order to make sure CI is happy let's simply comment out those assertions, we'll re-enable them as we fix root causes

Wdym? To comment out the assertions that lead to these two fails? Isn't it better to solve them or to even merge with the fuzzing action failure? The fuzzing itself seems to be working, but these fails should be fixed.

**Comment by alejandro-vaz** [comment]
since it's fuzzing I'll let it pass, but I don't think it's good to merge things that fail CI

can you rebase it??

**Comment by pedrodesu** [comment]
I agree with that generally. I think this is a good exception as it's something we just (re)introduced and that is apparently working, as the failing action is just a byproduct.

I can't work on it right now, I should be able to rebase it and perhaps investigate the remaining problem by tomorrow.

**Comment by alejandro-vaz** [comment]
I'm going to try to resolve conflicts myself

**Review by alejandro-vaz — APPROVED** [review]

**Comment by alejandro-vaz** [comment]
I managed to do it

thanks @pedrodesu for this new fuzzer, it is way better than what we had

**Comment by pedrodesu** [comment]
And indeed @alejandro-vaz, I also found the problem of the invariant failure. The problem appears from `Op::FromVec`:
```rs
small_vec = SmallVec::from_vec(small_vec.into_vec());
```
This is a completely valid operation. `small_vec` has initially a `length` of 5 and `capacity` of `N` (in the test, `N = 7`). `small_vec.into_vec()` transforms it into a `Vec` with a `length` of 5 and a `capacity` of 5. Now here comes the crucial error: `SmallVec::from_vec` simply emits a `SmallVec` guaranteed to be on the heap with the same `capacity` as the original `Vec`. It doesn't check whether it's smaller than `N`. This leads to the resulting `SmallVec` having a `length` of 5, a `capacity` of 5, but a `N` of 7, which triggers the assert and shows an invariant violation. This, of course, is incorrect behaviour. The fuzzer is indeed doing its job perfectly. I will open a PR soon trying to fix all found problems.

